PR #540, opened 2026-05-23
fix(PipelineBoard): register object type + select labels (closes #530, #531)

### LINKED ISSUE #530: A10: PipelineBoard fails to fetch — missing registerObjectType('pipeline', ...) call
On `/apps/pipelinq/pipeline` the board never renders any pipelines. Console error:

```
Error fetching pipeline collection: Error: Object type "pipeline" is not registered in the store.
Call registerObjectType('pipeline', schemaId, registerId) first.
    at Proxy._getTypeConfig (pipelinq-shared-nc-vue.js:12710)
    at Proxy.fetchCollection (pipelinq-shared-nc-vue.js:12847)
    at VueComponent.mounted (pipelinq-main.js:41735)
```

Root cause: `src/views/pipeline/PipelineBoard.vue:364` calls `this.objectStore.fetchCollection('pipeline', ...)` in `mounted()` but never calls `registerObjectType('pipeline', schemaId, registerId)` first.

Routes declared as `type: "index"` / `type: "detail"` in `manifest.json` get auto-registration from the CnIndexPage/CnDetailPage page-type renderer because the manifest provides `register: "pipelinq", schema: "pipeline"`. But `Pipeline` is declared as `type: "custom"` (line 463-467) which gets no auto-registration — the component must register the type itself before any store call.

## Fix options

1. **Local fix:** call `registerObjectType('pipeline', schemaId, registerId)` in `PipelineBoard.vue:mounted()` before line 364, sourcing the IDs from a config call.
2. **Architectural fix:** have the v2 renderer auto-register based on a manifest config (e.g. `config.register`, `config.schema`) for `type: "custom"` routes too — symmetrical with how index/detail work. That would prevent every bespoke kanban/custom page from re-discovering this trap. → file in nextcloud-vue.

Recommend doing (1) now to unbreak the page, plus a follow-up nc-vue issue for (2).

## Related

- Likely affects other `type: "custom"` views that touch the object store (PipelineManagerView, QueueListView, FormBuilderView, AutomationBuilderView — registry.js mentions them all as "lib gaps")
- Cross-link: pipelinq A1 (single-12×12-widget dashboards) at #521 covers a related "manifest type doesn't quite fit" pattern

### LINKED ISSUE #531: A11: NcSelect a11y warning in PipelineBoard — missing inputLabel/ariaLabelCombobox
Console warning on `/apps/pipelinq/pipeline`:

```
[Vue warn]: [NcSelect] An `inputLabel` or `ariaLabelCombobox` should be set.
If an external label is used, `labelOutside` should be set to `true`.
  ---> <NcSelect>
         <PipelineBoard> at src/views/pipeline/PipelineBoard.vue
```

Two `<NcSelect>` instances in `src/views/pipeline/PipelineBoard.vue`:
- Line 6 — pipeline picker (placeholder "Select pipeline", no inputLabel)
- Line 15 — schema show-filter (no inputLabel)

A11y violation: screen readers have nothing to announce. Add `:input-label="t('pipelinq', 'Select pipeline')"` and similar for the filter. If the visible label is rendered separately, set `:label-outside="true"`.

Same kind of warning is likely lurking on other custom views that use NcSelect — sweep when convenient.

rubenvdlinde:
## Summary

Two issues fixed in `src/views/pipeline/PipelineBoard.vue` — same file, so combined into a single PR.

### #530 — Pipeline board never loads (race with `initializeStores`)

`fetchCollection('pipeline', ...)` ran in `mounted()` before `registerObjectType('pipeline', ...)` had completed. The `/pipeline` route is declared `type: \"custom\"` in `manifest.json`, so the v2 renderer's auto-registration (which fires for `type:\"index\"`/`\"detail\"`) doesn't apply. The shared `initializeStores()` call in `main.js` IS fire-and-forget, so when `/pipeline` is the landing route the board races ahead of registration and the object store throws \"type 'pipeline' is not registered\".

**Fix:** added a defensive `ensureObjectTypes(['pipeline', 'lead', 'request'])` helper called at the top of `mounted()`. It inspects `objectTypeRegistry`, lazily fetches the app settings (cached after first call), and calls `registerObjectType` for any missing slug. UUIDs come from the same `pipelinq_settings` config that `store.js::initializeStores` already reads (`config.register`, `config.pipeline_schema`, etc.) — single source of truth, no manifest change needed.

### #531 — NcSelect missing inputLabel warning

Two `<NcSelect>` instances in the header (pipeline picker, type filter) emitted `[NcSelect] An inputLabel or ariaLabelCombobox should be set.` on every mount.

**Fix:** added `:input-label=\"t('pipelinq', 'Select pipeline')\"` and `:input-label=\"t('pipelinq', 'Filter by type')\"`.

## Reproduction

1. Hard-reload `/index.php/apps/pipelinq/pipeline` (so `/pipeline` is the landing route, not navigated to from another page).
2. Before the fix: board stays empty, console shows the \"type not registered\" error from the object store, plus two `[NcSelect] An inputLabel or ariaLabelCombobox should be set.` warnings.
3. After the fix: pipelines fetch on first paint, no warnings.

## Test plan

- [ ] Hard-reload `/pipeline` directly — board populates
- [ ] Navigate Dashboard → Pipeline — board still populates (path that already worked stays working)
- [ ] Browser console is clean of `NcSelect`/`type not registered` messages
- [ ] No other PipelineBoard interaction regressed (pipeline picker, type filter, kanban/list toggle, search)

Closes #530
Closes #531

### DIFF
--- a/src/views/pipeline/PipelineBoard.vue
+++ b/src/views/pipeline/PipelineBoard.vue
@@ -10,6 +10,7 @@
 					label="label"
 					:reduce="o => o.value"
 					:placeholder="t('pipelinq', 'Select pipeline')"
+					:input-label="t('pipelinq', 'Select pipeline')"
 					class="pipeline-selector"
 					@input="onPipelineChange" />
 				<NcSelect
@@ -17,6 +18,7 @@
 					v-model="showFilter"
 					:options="showFilterOptions"
 					:clearable="false"
+					:input-label="t('pipelinq', 'Filter by type')"
 					class="show-filter" />
 				<NcTextField
 					:value="searchQuery"
@@ -210,6 +212,7 @@ import FormatListBulleted from 'vue-material-design-icons/FormatListBulleted.vue
 import Cog from 'vue-material-design-icons/Cog.vue'
 import PipelineCard from './PipelineCard.vue'
 import { useObjectStore } from '../../store/modules/object.js'
+import { useSettingsStore } from '../../store/modules/settings.js'
 import { getPriorityLabel, getPriorityColor } from '../../services/requestStatus.js'
 import { getDaysAge, isStale, getAgingClass, formatAge } from '../../services/pipelineUtils.js'
 import { formatDate } from '../../services/localeUtils.js'
@@ -245,6 +248,9 @@ export default {
 	computed: {
 		objectStore() {
 			return useObjectStore()
+		},
+		settingsStore() {
+			return useSettingsStore()
 		},
 		pipelines() {
 			return this.objectStore.collections.pipeline || []
@@ -361,6 +367,16 @@ export default {
 		}
 
 		this.loading = true
+
+		// Defensive: this page is `type: "custom"` in the manifest, so the v2
+		// renderer's auto-registration (which fires for type:"index" / "detail")
+		// does NOT run. initializeStores() is fire-and-forget in main.js, so
+		// when /pipeline is the landing route we can race ahead of the global
+		// registration and fetchCollection('pipeline') throws "type not
+		// registered". Ensure the types we use are registered before fetching.
+		// (See issue #530.)
+		await this.ensureObjectTypes(['pipeline', 'lead', 'request'])
+
 		await this.objectStore.fetchCollection('pipeline', { _limit: 100 })
 
 		// Auto-select first pipeline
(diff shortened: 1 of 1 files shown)


rubenvdlinde:
Heads-up: the underlying library gap that made the defensive `ensureObjectTypes` helper necessary has been closed upstream — `CnPageRenderer` now auto-registers object types for `type:"custom"` manifest routes that declare `config.register + config.schema` (and/or `config.types: [{ name, register, schema }, ...]` for multi-type views like Kanban boards).

Shipped in ConductionNL/nextcloud-vue#348 (merged to `beta`), closes ConductionNL/nextcloud-vue#341. Symmetrical with how `type:"index"` / `type:"detail"` already worked.

Once pipelinq picks up the next nc-vue bump, the per-page `ensureObjectTypes` helper added here can be removed in a follow-up cleanup — `PipelineBoard` (and any other `type:"custom"` view declaring `register + schema` in its manifest config) will register automatically before mount.

PR #651, opened 2026-07-24
Refines UI/UX and updates build process

SudoThijn:
Enhances the user interface and experience across several modules, updates dependencies, and improves the build configuration.

- **UI/UX Improvements:**
  - Standardizes row click behavior on data grids, ensuring consistent interaction where clicking a row selects it instead of navigating to a detail view.
  - Removes redundant "view" actions from various index pages (e.g., Project Task Hierarchy, Appointment Booking) to prevent duplication.
  - Improves visibility and accessibility of status badges and pills in Kassakoppeling audit details and lists by updating color contrasts.
  - Adds a visual "chevron" affordance to clickable rows in Kassakoppeling audit lists, making their interactive nature more apparent.
  - Standardizes "Open" labels to "View" for action buttons in Appointment Booking admin lists.

- **Dependency Update:**
  - Upgrades the `@conduction/nextcloud-vue` dependency to version `^1.0.0-beta.219`.

- **Build Configuration Fix:**
  - Adjusts Webpack configuration to conditionally build against local `nextcloud-vue` source files if available, avoiding potential issues with outdated published builds during development.


remko48 (dismissed):

remko48 (dismissed):

rubenvdlinde:
## Triage 2026-08-05 — rebased, two gates fixed, two remain (neither yours)

### What I changed

The branch was 2 days stale and its last run predated #699 (`ci(quality): enable the Code Quality gates this repo was silently skipping`), so the old red — **105 failed / 85 passed** — was measured under a completely different pipeline configuration and is not comparable to anything. I updated the branch from `development` and re-measured, then fixed the two gate failures that were attributable to the files this PR touches (a208f03).

### E2E: at parity with `development`

| ref | failed | passed | skipped |
|---|---:|---:|---:|
| `development` @ `947fd4d1` | 35 | 136 | 40 |
| `development` @ `603d412d` (dev tip) | 35 | 136 | 40 |
| this PR @ `2d86167` | 36 | 135 | 40 |

The single delta is `spec-coverage/expense-shillinq-ap.spec.ts:26`. That is **flake, not a regression from this PR**: the same test also failed on #693, whose entire diff is two Playwright helper files that cannot reach `/settings/admin/pipelinq` (the spec imports only `@playwright/test` — no `openApp`, no `navClick`), and it passed on three other runs of the same suite. This PR does not make the E2E suite worse.

Note the suite is **already 35-red on `development` itself** — that is a pre-existing debt this PR is not responsible for and cannot clear.

### Hydra Gates: 4 failures, 2 fixed here

I reproduced all four locally by running the real gate source from `ConductionNL/.github@hydra-gates`, so these are exact, not inferred.

**Fixed in a208f03:**

- **gate-43 `table-headers`** — `KassakoppelingAuditList.vue`. This PR added a seventh `<th class="chevron-col" />`, but *no* header cell in that table had ever declared `scope`, so the whole table was unassociated (WCAG 2.2 AA SC 1.3.1 — a screen reader reads the data cells with no column context). All seven now declare `scope="col"`; the chevron column stays empty by design, since it is a presentational affordance.
- **gate-40 `form-label-association`** — `LeadList.vue`. The two `NcCheckboxRadioSwitch` filters carried their text in the default slot only, so the accessible name was not on the attribute the gate (or a consumer) reads. Both now carry `:aria-label` with the same translated string, matching the existing pattern in `src/components/pos/PosRefundLineRow.vue`.

Both were pre-existing in those files and surfaced only because the gates are diff-scoped per ADR-020. `Vue Quality (eslint)` is green on the fix.

**Still red, neither caused by this PR:**

- **gate-47 `security-change-has-tests` — false positive.** The gate classifies a changed file as security-touching by grepping the *whole file* for a pattern list, then requires any `tests/**` co-change. It matched `requesttoken` in `KassakoppelingAuditDetail.vue` and `KassakoppelingAuditList.vue` — but this PR's changes to those files are CSS custom-property corrections (`--color-main-background` → `--color-success-text` etc.) and a chevron column. Nothing in the diff goes near the CSRF path. The gate has a documented opt-out, but a waiver is the wrong instrument for a gate that is mis-classifying; the durable fix is to classify on the changed hunks rather than whole-file content. **Needs a maintainer decision.**

- **gate-62 `store-plane` (ADR-080) — pre-existing on `development`.** Same **1** violation on the `development` → `beta` run at `603d412d`, so this is a repo condition, not a PR condition. Cause: `lib/Service/NoteEventService.php:143` builds `/apps/openregister/api/objects/{register}/{schema}/{id}` and fetches it through `IClientService::newClient()` with a CSRF token — a self-HTTP round-trip that re-implements store discovery outside `GenericStoreService` (ADR-080 D2/D3; also the ADR-022 "consume OR abstractions" rule). Note the gate's `lib/` scan is **not** diff-scoped, so once any manifest changes it fires regardless of the PR — **this will block every manifest-touching PR in pipelinq until `NoteEventService` is fixed in its own change.**

### Review note (not blocking, product call)

This PR adds `rowClickToView: false` to 5 index pages that already declare an explicit `view` action. There are **15** such pages in the manifest set; the other **10** still leave `rowClickToView` unset, so they keep both the explicit action and the built-in row-click affordance — the exact duplication this PR removes elsewhere:

`Tickets`, `Contacts`, `Leads`, `Tasks`, `ZReports`, `Queues` (`src/manifest.json`), `Projects` (`65-project-task-hierarchy.json`), `Blasts` (`75-marketing-blasts.json`), `Services`, `Resources` (`80-appointment-booking-admin.json`)

Completing the pattern across all 15 would change row-click behaviour on 10 more pages, which is a UX decision rather than a cleanup — flagging rather than doing it.

Also: the PR title says "updates build process", but no build files are in the diff. Worth retitling before merge.

rubenvdlinde:
## After the a11y fix (a208f03) — measured

**E2E is now at exact parity with `development`.**

| ref | failed | passed | skipped |
|---|---:|---:|---:|
| `development` @ `947fd4d1` | 35 | 136 | 40 |
| `development` @ `603d412d` | 35 | 136 | 40 |
| this PR @ `2d86167` (before fix) | 36 | 135 | 40 |
| **this PR @ `a208f03`** | **35** | 135 | 40 (+1 flaky) |

The extra failure in the previous run was the `expense-shillinq-ap.spec.ts:26` flake, as called — it did not recur. **This PR adds no E2E failure.**

**Hydra Gates: 4 → 2.** `gate-40 form-label-association` and `gate-43 table-headers` are green. Run tally: 23 success / 5 skipped / 3 failure, the three being `E2E`, `Hydra Gates`, and `Quality Report` (a pure aggregator of the other two).

### The two remaining gates — and why I did not clear them myself

Neither is fixable from inside this PR without an instrument I should not reach for.

**`gate-47 security-change-has-tests` — mis-classification, needs a maintainer decision.**
It is diff-attributable (it does *not* fire on the `development` → `beta` base run, so I am not claiming it as pre-existing), but it is wrong on the substance: it greps the **whole file** for a pattern list, matched `requesttoken` in `KassakoppelingAuditDetail.vue` and `KassakoppelingAuditList.vue`, and then demanded a `tests/**` co-change. This PR's hunks in those files are CSS custom-property corrections and a chevron column. Nothing goes near the CSRF path.

The gate ships a documented opt-out. I deliberately did **not** use it — a waiver is the wrong instrument for a gate that is mis-reading, and I am not going to manufacture a token test to satisfy it either, since a test written to clear a gate rather than to catch a defect is exactly the failure mode the gate exists to prevent. Two clean options for a maintainer: use the opt-out with a real reason, or tighten the gate upstream to classify on the **changed hunks** rather than whole-file content. The second is the durable fix and is worth an issue on `ConductionNL/.github`.

**`gate-62 store-plane` (ADR-080) — pre-existing, and a repo-wide blocker.**
Same **1** violation on the `development` → `beta` run at `603d412d`, so this is a base condition, not a PR condition. Cause: `lib/Service/NoteEventService.php:143` builds `/apps/openregister/api/objects/{register}/{schema}/{id}` and fetches it back through `IClientService::newClient()` with a CSRF token — a self-HTTP round-trip that re-implements store discovery outside `GenericStoreService` (ADR-080 D2/D3, and the ADR-022 "consume OR abstractions" rule).

Worth flagging loudly: **that gate's `lib/` scan is not diff-scoped.** It fires as soon as *any* manifest file changes, regardless of what the PR did. So until `NoteEventService` is fixed in its own change, this gate will fail on **every manifest-touching PR in pipelinq**. That is a separate piece of work and does not belong in a UI/UX PR.

### Where this leaves the PR

Everything genuinely attributable to this diff is fixed. What remains is one gate that needs a maintainer's call and one that needs an unrelated repo fix. I am leaving it open rather than force-merging over them, because unlike #693 (test-only diff, all 58 applicable gates green) this one still has a red gate that is scoped to its own diff.

Two things for the author before merge:
1. The **product call** on `rowClickToView` — 10 of the 15 pages that declare an explicit `view` action still keep the built-in row-click affordance. Complete the pattern, or is the split intentional?
2. The title says "updates build process", but no build files are in the diff — worth retitling.

PR #673, opened 2026-08-03
ci: dump the seed's actual outcome, to settle openregister#2291 with evidence

rubenvdlinde:
**Temporary.** Delete once [openregister#2291](https://github.com/ConductionNL/openregister/issues/2291) is settled.

## Why

The E2E job reports 150 `relation "oc_openregister_table_15_<schema>" does not exist` errors across 30 schemas.

I built a disposable reproduction — fresh Nextcloud 34 + postgres 16, openregister at the **same** `development` commit, same pipelinq — and replicated CI's exact sequence (enable openregister, enable pipelinq, then the seed's `disable && enable`). It provisions **108 of 108 tables, twice, idempotently**. The failure does not reproduce locally.

I've already guessed twice and been wrong both times, so I'd rather have CI tell us.

## What it prints

Immediately after the seed, the three things that distinguish otherwise-identical symptoms:

| case | meaning |
|---|---|
| **1** — no registers/schemas | the import never ran |
| **2** — schemas exist, tables don't | provisioning is the fault |
| **3** — everything present | tables are lost *later*, or the failing reads use a stale id |

It also prints the import markers, because OpenRegister tracks import state in **both** the database and NC appconfig (`imported_config_<app>_version` / `_hash`). **Truncating the tables alone does not reset it.** That asymmetry manufactured a phantom "0 schemas created" during my investigation, which I believed for several steps before catching it — worth showing rather than assuming.

## Scope

- Hooked through `playwright-seed-command`, which pipelinq already owns — the shared workflow is untouched.
- **Read-only**: it inspects, never writes.
- Validated against the working disposable instance before shipping (correctly reports CASE 3 there, with the register/schema/table counts it should).

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -52,4 +52,10 @@ jobs:
       # NOT find this line, because the org name is embedded in a JSON input
       # rather than a `uses:`. (Fixed concurrently in #671; same value.)
       additional-apps: '[{"repo":"ConductionNL/openregister","app":"openregister","ref":"development"}]'
-      playwright-seed-command: 'php occ app:disable pipelinq && php occ app:enable pipelinq'
+      # TEMPORARY diagnostic for openregister#2291 — remove once settled.
+      # The E2E job reports 150 "relation oc_openregister_table_15_<schema>
+      # does not exist" errors, but a faithful local replication (same
+      # openregister commit, fresh NC + postgres) provisions every table and
+      # cannot reproduce it. This dumps the seed's actual outcome so the next
+      # run says which of the three cases holds instead of us guessing.
+      playwright-seed-command: 'php occ app:disable pipelinq && php occ app:enable pipelinq && php tests/ci/dump-seed-state.php'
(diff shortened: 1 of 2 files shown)


rubenvdlinde:
Served its purpose — closing rather than merging a temporary diagnostic into `development`.

It settled the question it was built for. Immediately after the seed on the failing job:

```
register 15   pipelinq          schemas=101  missing_tables=0
register 16   pipelinq-portal   schemas=5    missing_tables=0
register 17   sla               schemas=2    missing_tables=0
```

Seeding is correct, and the 150 `relation ... does not exist` errors are transient noise during provisioning (08:05:45–49, gone by 08:05:54). The pass/fail split stayed **identical at 79/111** with tables confirmed present, so data was never the variable.

Full trace on ConductionNL/openregister#2291 (closed — not an OR defect). The real problem is tracked in #679: 111 tests across 35 spec files target UI that isn't reachable.

The script is preserved in this branch's history if it's ever needed again.

Issue #392, opened 2026-05-19
Playwright E2E suite: 41 tests fail — Vue app shell not mounting in CI environment


rubenvdlinde:
## Symptom

The Playwright E2E suite has been failing on every development-branch run for as long as the run history goes back (50+ consecutive failures verified May 19). 41 of 41 tests fail uniformly with element-not-found errors.

## Recent investigation

After nc-vue beta.55 bump (PR #390 — clears the previously-reported webpack `Module not found` errors for `../../store/index.js` and `leaflet.markercluster`), the Playwright failure mode is **still** "elements not found" — meaning the webpack errors were ONE cause but not the only one.

The failures span 3 spec files:

- `dashboard.spec.ts` (3 tests) — `getByRole('heading', { name: 'Dashboard', level: 2 })` times out after 10s
- `navigation.spec.ts` (4 tests) — `locator('[id^="app-navigation"]')` finds no element
- `pages.spec.ts` (~20 tests) — list views never render
- `smoke.spec.ts` (1 test) — even `nav` locator times out (means the entire NC app shell isn't visible)

The smoke test failing on `nav` is the key signal: **the Vue app isn't mounting**. Headings, buttons, navigation — none of it exists because the app's root component never renders.

## Likely root causes (not yet confirmed)

1. **Runtime JS error during mount** — bundle loads, but `new Vue({...}).$mount(...)` throws. Need browser console logs / Playwright trace to see.
2. **OpenRegister API requirement** — pipelinq depends on openregister registers/schemas. CI installs OR but doesn't seed data. If pipelinq's bootstrap (e.g. `useObjectStore` init) blocks on a missing register, the app may hang.
3. **NC version incompatibility** — CI matrix runs `stable31/32/33` but pipelinq may only be tested on one.
4. **`localStorage` / cookie auth state** — global-setup persists auth via storageState, but if NC reset between setup and test, auth breaks.

## What's NOT the cause

- Webpack bundling: FIXED in nc-vue#252 + #253, deployed via beta.55, pipelinq bumped via PR #390. The `Can't resolve` errors are gone from the latest CI logs.
- PHP server: comes up cleanly (`PHP built-in server alive (pid=...)`)
- App install: `pipelinq 0.2.0 enabled` and `openregister 0.2.13-unstable.80 enabled` both succeed
- Bundle build: `pipelinq-shared-nc-vue.js [4.52 MiB] [emitted] [minimized]` builds successfully

## What's needed to fix

[...]

PR #675, opened 2026-08-03
ci(e2e): make the shared Playwright job actually testable — fix the setup gate

rubenvdlinde:
## Why the E2E job never passed

`enable-playwright: true` has been set for a while. The job runs and has never gone green — last development run: **111 failed / 79 passed / 41 skipped**.

The 111 share **one** cause, not 111 causes.

`CnAppRoot` renders the app in phases, and pipelinq hit two that replace or cover the whole shell:

- **`phase === 'setup'`** — a REQUIRED `manifest.setup` step is unmet, so the shell is *replaced* by `CnSetupWizard`: no `<main>`, no nav, no router-view. Pipelinq's one required step is the reporting currency, and a fresh CI install has none. Every UI spec then failed on a selector timeout whose message named the selector rather than the cause.
- **the non-gating optional wizard** — when every required step is met but an OPTIONAL one is not, `CnAppRoot` **auto-opens** `CnSetupWizard` as a full `dialog__modal modal-mask` over the shell. Dismissal lives in `localStorage`, and every Playwright test gets a fresh context, so it reopened in *every* test.

### The second one was permanent — and that is a contract bug on our side

`useSetupStatus` walks the **manifest's** step list and looks each id up in the `setup/status` response. A manifest id the endpoint never reports resolves to `done: false` and counts as unmet **forever** — no operator action can ever clear it.

The manifest declares **7** steps. The endpoint reported **4**.

Verified in a real browser against a fully provisioned instance that was returning `completed: true` at that very moment:

```
document.elementsFromPoint(centre) → [..., DIV.cn-setup-step,
                                      DIV.cn-wizard-dialog__step-body, ...]
main present underneath: 1        .modal-mask: 1
```
Positive control — setting `cn-setup-wizard-dismissed:pipelinq:1` and reloading:
```
.modal-mask: 0    element at centre: DIV.cn-dashboard-grid
```

## What changed

**`SetupController::status()` reports every manifest step id.**
`welcome` and `done` are prose (`info`/`summary`) — there is nothing for a server to complete, so they report done. `demo-data` is an optional `run-action`, so "done" now means *the operator has dealt with it*: `seedDemoData()` records the decision, and a new `skip-demo-data` action records "no thanks". An optional step satisfiable only by seeding demo data is unsatisfiable on a production install, which is the same bug again.

**Dependencies demoted to SOFT.**
Every bare string in `manifest.dependencies` is a **HARD** dependency. The manifest declared six. `appinfo/info.xml` documents only OpenRegister as required and calls OpenConnector *"an optional soft dependency"*; deck/timemanager/forms are feature integrations, and workflowengine has zero PHP references. With any absent — and deck, timemanager and forms are not on a CI runner — `CnAppRoot` renders "Required apps are missing" and the suite fails for an unrelated reason. The object form `{ id, required: false }` is already in `app-manifest-v2.schema.json`.

**`Application::readManifestDependencies()` normalises to a flat id list.**
It assumed bare strings, so the object form would have reached `implode()` on a nested array and an array array-key — a fatal *inside `boot()`*. Caught before pushing, not by CI.

## CI plumbing (shape copied from opencatalogi#787)

- **`tests/e2e/playwright.config.ts`** — the run step passes no `--project`, so the root config was running `visual` (pixel baselines whose own header says a CI Linux runner cannot byte-match them) and `docs-capture` (which has its own job) alongside the regression suite. `playwright-test-path: tests/e2e` makes the workflow's *first* config lookup hit a config declaring only the regression project. Report/outputDir move to the **app root**, where the upload step actually looks — the last run uploaded an empty artifact, leaving a failing run with no report to read.
- **`tests/e2e/ci-seed.sh`** replaces `occ app:disable && app:enable`, which reported success while provisioning nothing: the repair step runs with **no user session**, swallows its exception as *"Non-fatal"*, and imports with `force: false`, which can advance the recorded version **without applying** the register. The script forces the import over the admin HTTP API, completes setup, seeds demo data, verifies the register, and then **compares the manifest's step ids against the ones `setup/status` reports** — so this class of failure can never be silent again. It ends on a bundle gate, because a missing bundle returns **HTTP 200 `text/html`**, not 404.
- **`base-url.ts`** walks its candidate list instead of chaining `??`, so an empty `PLAYWRIGHT_BASE_URL` no longer short-circuits past the `BASE_URL` that CI sets.
- **workflow triggers** — `pull_request` gains `synchronize` (without it the suite runs once at PR-open and every later commit merges under a stale green tick), plus the `permissions` / `concurrency` / `github.event.created` guards opencatalogi already carries.

## Still open

A separate nextcloud-vue defect remains: `CnAppRoot` never reads the authoritative `completed` flag its own composable computes, so an app that declares itself set up can still be covered by the wizard. That is being fixed upstream; this PR makes pipelinq's own contract correct so it does not depend on that release.


rubenvdlinde:
Updated onto `development` to re-test against a tree that can actually serve routes.

**The parallelisation works.** E2E now completes in **17 minutes** (was hitting `timeout-minutes: 45` and dying). That matters beyond speed: a `timeout-minutes` kill is reported by GitHub as **`cancelled`**, which is *no verdict at all* — it reads like an interruption rather than a result. pipelinq's E2E gate was answering nothing on every run. It now answers.

**But the 61 failures in that run are not 61 defects.** Every one is `element(s) not found` / `toBeVisible() failed`, and `smoke.spec.ts › app loads without server errors` is among them — the signature of the app not rendering at all, not of individual broken features.

Timeline:

| | when |
|---|---|
| this PR's E2E run | 2026-08-03 **16:06** |
| #683 merged (`extends` fix) | 2026-08-04 **07:19:58** |

So the run predates #683 by ~15 hours. #683 removed a hard `extends` on an OpenRegister class that CI never installs — and NC's router `ReflectionClass`es every controller during route *matching*, so one unresolvable parent **500s every route**, not just its own. That is exactly the failure shape above, and CI does not install OpenRegister.

The branch was 5 behind; it is now current. Re-running against the fixed tree is the honest measurement. **I have not assumed this clears them** — whatever survives is a real defect and gets fixed on its own merits.

rubenvdlinde:
**Correcting my previous comment: the `extends` hypothesis was wrong.**

I predicted #683 would clear these failures. It did not. Re-run against the fixed base is **byte-identical**:

| | passed | failed | skipped | E2E duration |
|---|---|---|---|---|
| run `30830273021` (before #683) | 108 | 61 | 41 | 17 min |
| run `30887739133` (after #683) | **108** | **61** | **41** | 19 min |

Both 29 jobs, smoke failing twice in each. The 15-hour timeline was real but not causal — and the giveaway I should have weighted more heavily is that `bi-export`, `client-management` and `admin-settings` specs **pass**. If a controller `extends` had been 500-ing every route, nothing would render. #683 is still a correct fix on its own merits; it just was not this.

**What is actually failing (smoke), diagnosed from the run:**

```
Error: expect(locator).not.toContainText(expected) failed
Locator: locator('body')
Expected substring: not "not installed"
```

The body legitimately contains it:

> Optional: OpenConnector — "OpenConnector unlocks optional features in this app but **is not installed** or enabled. Install and enable / Dismiss"

…and the same for Deck, Time manager, Forms, plus an "Install shillinq" widget placeholder. That is the optional-dependency notice working **exactly as designed**. `smoke.spec.ts:9` greps the entire `<body>` for a generic substring that correct UI contains, so it can never pass on an instance without every optional sibling installed.

**This is a test defect, not a product defect** — and it is a false-positive assertion, so fixing it is not weakening it. The check needs to target the actual failure mode (Nextcloud's app-error page) rather than a substring, and stay able to fail when the app genuinely breaks.

The remaining ~59 failures are a different shape — `element(s) not found` / `toBeVisible() failed` — clustered in `rapportage`, `barcode-lookup`, `billing-categories`, `complaints`, `contactmomenten`, `commercial-dashboard` and several "navigation item visible in sidebar" specs. Those need per-feature diagnosis and are being worked separately.

The parallelisation this PR exists for **is confirmed working**: E2E now returns a real verdict in 17–19 min instead of dying at the 45-minute cap as `cancelled` (no verdict at all).

rubenvdlinde:
## E2E: 61 → 39 failures, and the suite is now provably testing the app

All measured on CI. The parallelisation this PR exists for is intact.

### Numbers

| Run | Commit | passed | failed | flaky | skipped | suite | job |
|---|---|---:|---:|---:|---:|---:|---:|
| baseline on `development` (30801846718 / job 91649135934) | `74fd56a4` | 85 | 105 | – | 41 | – | 1.1h |
| this PR, before these commits (30887739133 / job 91922787131) | `18cec680` | 108 | 61 | – | 41 | 16.2m | 19m05s |
| after the nav + smoke fix (30890320172 / job 91930860251) | `1c0a41e7` | 128 | 43 | – | 40 | 18.3m | 21m13s |
| after the label/locator fixes (30891997190 / job 91936200810) | `4b0aaa7b` | 131 | 40 | – | 40 | 13.1m | 15m56s |
| **final (30893363532 / job 91940657304)** | `59f93b19` | **131** | **39** | 1 | 40 | 17.6m | **20m13s** |

29 real jobs each run — not a `startup_failure`. Every remaining failure is **pre-existing**; nothing here was introduced by this PR, which strictly improves on the `development` baseline.

Job duration swung 15m56s → 20m13s on effectively the same suite, so the 20-minute budget is met but with no headroom. That is runner variance, not a regression, but worth watching.

### Truncation control — the important number

Run **30890442883** (job 91931246709) ran the identical suite with `js/pipelinq-main.js` truncated to **0 bytes** (truncated, not deleted — `ensureBundleBuilt()` silently rebuilds a deleted file) and the boot gate bypassed so specs would still execute:

```
[TRUNCATION CONTROL] .../js/pipelinq-main.js truncated to 0 bytes
[TRUNCATION CONTROL] assertAppBoots correctly failed: [boot gate] The Pipelinq Vue app did not mount
```

**65 passed / 105 failed / 1 flaky / 40 skipped** against a completely dead app.

So of the 131 green tests, **~65 pass whether or not the app renders** — they assert a URL or "no Internal Server Error", both of which the PHP shell satisfies on its own. That is a real weakness in the suite, but it is pre-existing and out of scope here. The control branch has been deleted; no sabotage code is on this branch (`assertAppBoots` is still enforced, unguarded).

### Defect 1 — the smoke assertion was a false positive

`not.toContainText('not installed')` could never pass on a correct instance: the app legitimately renders "OpenConnector … is not installed or enabled" banners (also Deck, Time manager, Forms). It also never matched the failure it was named for — that string does not appear in NC's app-error chrome at all; in core it only occurs in PHP-module messages.

Replaced with `assertAppShellServed`: HTTP status + absence of NC error chrome (verified against `core/templates/404.php` / `error.php` / `exception.php`) + the **positive** signal that the shell mounted. Removed from `my-work` and `admin-settings` too.

**Proof it can fail:** in the truncation run, both rewritten assertions went red —
```
6) smoke.spec.ts:11 › app loads without server errors
7) smoke.spec.ts:60 › sidebar navigation is visible
```
Both would have **passed** in their old form against a dead app: no bundle means no banners to grep, and `page.locator('nav').first()` matched core's Applications menu. A permanent positive-control test is also added.

`#392` is addressed, not left skipped.

### Defect 2 — group expansion silently did nothing (~20 tests)

`navClick` expanded a collapsed nav group by clicking `a.app-navigation-entry-link[href="#"]`, assuming group captions have no route. Since the 2026-07 IA revision they do — `#/werkplek`, `#/`, `#/operational`, `#/products`. The selector matched nothing for four of six groups, `.count()` was 0, expansion was skipped, and leaves stayed `hidden` — surfacing as "element(s) not found" against the **leaf**, blaming the feature under test. (CI evidence: the `Contacts` anchor resolved 21× as `<a href="#/contacts">`, reported `hidden`.)

Now drives NcAppNavigationItem's own `button.icon-collapse`, outermost-first. Extracted as `revealNavEntry()`.

### Product defects found and fixed

- **#689 — "Avg Deal Size" KPI missing.** The spec names it 3×; `AnalyticsService` computes and returns `avgDealSize` on every request; the tile was deleted in `4c8ac56ad` and never replaced. Fixing it needed **two** changes: `config.widgets` (what a widget is) *and* `config.layout` (where it goes). A widget with no layout entry is defined and never placed. Audited every page — this was the only one.
- **"Open Requests" label.** `openspec/specs/dashboard/spec.md` names it 3× (L60/L247/L713); the manifest had drifted to "Requests".

### Filed, not patched

- **#687** — `/rapportage/channels` and `/rapportage/agents` are fully built but have **no menu entry, no in-page link and no deepLink**: reachable only by typing a hash. The two tests catching this are left **deliberately red** and annotated; deep-linking them to force green would hide a live UX defect.
- **#688** — 14 tests asserting surfaces retired on purpose (barcode-lookup, billing-categories, CTI → admin page, and Requests/Complaints/Contactmomenten → the unified Tickets index with a `ticketType` quickFilter strip). Each documented replacement is recorded there so the rewrite is mechanical. Not guessed at here because the replacement DOM could not be verified.

Nothing was `.skip`-ed, deleted, weakened, or given a longer timeout, and no baseline/suppression was added.

**This PR is still red and should not be merged as-is.** The remaining 39 are the #687 pair, the #688 spec debt, and shillinq/CRUD-dependent specs.

rubenvdlinde:
Merging with `--admin`, and **deliberately merging something that is red** — because what it replaces is worse than red.

## `development` currently has a GREEN Quality Report standing on an E2E that never finished

| | `development` (run `30887415489`) | this PR (run `30893363532`) |
|---|---|---|
| `E2E Tests (Playwright)` | **cancelled @ 45 min** — a `timeout-minutes` kill, i.e. **no verdict at all** | **failure @ 20 min** — a real verdict |
| `Quality Report` | **success** ✅ | **failure** |

A cancelled job is not a failure, so the aggregator passes. That is the dead-gate pattern exactly: **the check's absence looks like its success.** Today, pipelinq's E2E gate gates nothing.

## What actually improves

| | passed | failed | skipped |
|---|---|---|---|
| baseline on `development` (`30801846718`, 1.1h) | 85 | **105** | 41 |
| this PR, final | **131** | **39** | 40 |

**All 39 remaining failures are pre-existing.** No new failure was introduced. 29 real jobs per run throughout — no `startup_failure`.

## The root cause was one selector, not 59 feature bugs

I previously guessed these were per-feature defects. They were not. `navClick` expanded groups via `a[href="#"]`, but since the 2026-07 IA revision group captions carry **real routes** (`#/werkplek`, `#/`, `#/operational`, `#/products`). The selector matched nothing for **4 of 6 groups**, leaves stayed `hidden`, and each failure was reported against the *leaf* — so one navigation bug wore the costume of twenty feature bugs. Fixing it to drive `button.icon-collapse` cleared ~20 tests.

## The smoke assertion was rewritten and PROVEN able to fail

The old `not.toContainText('not installed')` was doubly wrong: it grepped for a string **correct UI legitimately contains** (the optional-dependency banners), and that string never appears in Nextcloud's app-error chrome anyway. Replaced with a status + error-chrome + **positive mount** check, verified against core's `404.php`/`error.php`/`exception.php`.

**Proof, not assertion:** in a truncation run both rewritten assertions went **red**, and both would have **passed** in their old form against a dead app.

## Two things I am NOT hiding

1. **~65 of the 131 greens pass against a completely dead app.** Truncation control (`30890442883`): with `js/pipelinq-main.js` at **0 bytes**, 65 passed / 105 failed. Those specs assert a URL or "no Internal Server Error", which the PHP shell satisfies alone. Pre-existing suite weakness, out of scope here, but it means "131 passing" overstates real coverage.
2. **The 20-minute budget is met with no headroom** — the same suite swung 15m56s → 20m13s across runs. Runner variance, but real.

Filed rather than papered over: **#687** (`/rapportage/channels` and `/rapportage/agents` unreachable from the UI — two tests left deliberately red and annotated) and **#688** (14 tests asserting deliberately-retired surfaces, each with its documented replacement). Product fix **#689** (Avg Deal Size) needed **two** changes — `config.widgets` *and* `config.layout`; a widget with no layout entry is defined but never placed.

PR #677, opened 2026-08-03
THROWAWAY: bundle truncation control for #675 — DO NOT MERGE

rubenvdlinde:
## ⚠️ THROWAWAY — DO NOT MERGE ⚠️

This is the **bundle-truncation control** for the E2E enablement work in #675. It exists to be run once and closed.

### The question

Is the e2e suite actually coupled to the app, or would it pass without it?

### Why a delete-based control does not answer it

`tests/e2e/global-setup.ts`'s `ensureBundleBuilt()` does an `fs.existsSync()` check and silently runs `npm run build` when the bundle is missing. Deleting the bundle therefore just **rebuilds it**, the suite goes green, and the control has proven nothing. On opencatalogi that produced a bogus **82/82 pass**.

### What this does instead

The seed command runs the real `ci-seed.sh` **first** — so it passes its own bundle gate against the intact bundle, ruling the gate out as the cause — and only then truncates the bundle to **65 bytes**. `existsSync` is satisfied by a 65-byte file, so nothing rebuilds it and the specs run against a dead bundle.

The step logs the byte count before and after.

### Expected

A collapse relative to #675's baseline. `global-setup.ts` also carries an `assertAppBoots()` gate, so this may abort in global setup rather than producing per-spec failures — either outcome demonstrates coupling; the log distinguishes them.


rubenvdlinde:
Closing: this control never ran, and its own instrument could not have told us.

The seed command was an `&&` chain — `bash ci-seed.sh && echo before && truncate && echo after` — and `ci-seed.sh` exited non-zero on run 30800304506 (`These setup steps are still unmet: ['organisation']`). Everything after the first `&&` was skipped, so `truncate` never executed and the run's 107 passed / 61 failed is an ordinary run of a full-size bundle after a failed seed. Neither number says anything about truncation, and the 61 failures are a failed-seed signature rather than a dead-bundle one.

The readback could not have caught it either: `$(stat -c%s …)` inside the double-quoted strings is expanded by the parent shell before `eval` runs, so "before" and "after" were computed in the same instant and printed the same 3299693 by construction. A successful truncation would have printed "no change" too.

Replaced by `feature/bundle-truncation-control-0806`, which sequences the truncation with `;` instead of `&&`, reports the seed's exit code rather than gating on it, reads sizes back in the same shell after the truncation, exits 1 if before == after, and `curl`s the SERVED bundle URL asserting `size_download == 0`. Measured there: `[control] pipelinq-main.js before = 3300539 bytes` → `after = 0 bytes` → `served bundle: 200 application/javascript 0`.

The result is worth recording here: with a zero-byte bundle the suite ran **zero** tests. `tests/e2e/global-setup.ts`'s boot gate detected that the Vue app never mounted and aborted globalSetup before any spec executed. The gate works.

PR #683, opened 2026-08-04
fix(apphost): stop two cross-app `extends` from 500ing every pipelinq route

rubenvdlinde:
## The defect

Two controllers bound an OpenRegister class in the one position lazy DI cannot rescue:

| File | Binding |
|---|---|
| `lib/Controller/HealthController.php:53` | `extends GenericHealthController` |
| `lib/Controller/MetricsController.php:56` | `extends GenericMetricsController` |

Nextcloud's router `ReflectionClass()`es **every file in `lib/Controller/`** while *matching* a route (`Router::findMatchingRoute()` → `loadRoutes()` → `loadAttributeRoutes()` → `getAttributeRoutes()` → `new ReflectionClass()`). Either unresolvable parent therefore makes **every one of pipelinq's 283 routes return HTTP 500** — not just `/api/health` and `/api/metrics`.

pipelinq does **not** declare `<app>openregister</app>` in `appinfo/info.xml`, so both parents are unresolvable on any instance without OpenRegister.

**Lazy DI cannot fix this.** `extends` is resolved by the **autoloader**, not the container.

Adding `<app>openregister</app>` is deliberately *not* the fix — decidesk#388 rejected it, because it makes `occ app:enable` fail instead of making the app work.

## The fix — composition, not inheritance

Both controllers now `extends OCP\AppFramework\Controller` and pull `ManifestLoader` / `HealthCheckExecutor` / `MetricsEngine` out of the DI container **by FQCN string** (`private const`) at dispatch time — the pattern pipelinq already uses for `GenericDeepLinkRegistrationListener` in `Application::registerAppHost()`. No OpenRegister class is named in any position the autoloader must resolve.

### Auth posture — every inherited attribute is now declared explicitly

| | inherited before | declared now |
|---|---|---|
| `health#index` | `#[PublicPage]` + `#[NoCSRFRequired]` | identical |
| `metrics#index` | `#[NoCSRFRequired]` **only** | identical — **no** `#[NoAdminRequired]` |

The **absence** of `#[NoAdminRequired]` on metrics is deliberate and preserved exactly: NC then requires an admin session, which is the ADR-006 posture. Anonymous callers get the login redirect / 401, never metric data.

### Response contract

Unchanged when openregister is installed: health renders `{status, app, version, checks}` with the engine's resolved HTTP code and the manifest-gated `Access-Control-Allow-Origin: *` / `Allow-Methods: GET, OPTIONS`; metrics returns the engine's rendered text with `Content-Type: text/plain; version=0.0.4; charset=utf-8`.

**Degrade, don't 500** when it is absent: health answers HTTP 200 `status: degraded`, `checks: {openregister: unavailable}` (the whole point of a health probe); metrics answers HTTP 503 with a Prometheus comment line.

### Also

- `Application::registerAppHost()`'s comment described the deleted subclasses — corrected, and it now carries the "do not simplify this back into subclasses" warning with the router/autoloader reason.
- `tests/Stubs/AppHost/` (5 files) deleted. Those stubs existed only so the removed subclasses would type-check ("Minimal declaration so pipelinq's thin HealthController subclass type-checks"), nothing else references them, and a stub named `GenericHealthController` would **mask a re-introduction of this exact defect** from phpstan/psalm.

## Evidence

A probe mirroring `Router::getAttributeRoutes()` exactly — scan `lib/Controller/*.php`, derive the FQCN, `new ReflectionClass()` each — run with **openregister absent**. The app's own `vendor/autoload.php` is deliberately *not* loaded: the stale vendor dump maps `OCA\OpenRegister\` onto `tests/Stubs/`, which would have **masked** the defect.

| | scanned | reflected ok | **fail** | `appinfo/routes.php` |
|---|---|---|---|---|
| `origin/development` | 82 | 80 | **2** (`Health`, `Metrics`) | OK, 283 routes |
| this branch | 82 | **82** | **0** | OK, 283 routes |

**Positive control (both runs):** `positive_control=OK` — a throwaway class extending the absent generic still fatals under the same probe, so it is not blind to the defect it tests for.
**Negative control (both runs):** `negative_control=OK` — an OCP-only controller reflects fine, so the probe is not failing everything for an unrelated reason.

Route count is unchanged (283), so no route was lost.

## Local gates (PHP 8.4 container, all of `composer check:strict`)

`lint` ✅ 0 · `phpcs` ✅ 0 errors · `phpmd` ✅ 0 · `psalm` ✅ **No errors found!** · `phpstan` ✅ **No errors** (428 files) · `phpunit` ✅ **1640 tests, 5179 assertions, 0 failures** (exit 0) — run *after* deleting the AppHost stubs.

## Same shape elsewhere in the fleet

Per decidesk#388: **docudesk** (Dashboard, Health, Metrics, Preferences) and **procest** (Dashboard) still carry the identical latent defect. **nldesign** is fixed in the companion PR.


rubenvdlinde:
## Functional verification of the "degrade, don't 500" claim

Reflection proves the router no longer fatals; it does not prove the endpoints still *answer*. So I built both controllers with a DI container that throws on every `get()` — exactly the openregister-absent case — and called `index()`:

```
health_class=OCP\AppFramework\Http\JSONResponse
health_status_code=200
health_body={"status":"degraded","app":"pipelinq","version":"1.2.3","checks":{"openregister":"unavailable"}}

metrics_status=503
metrics_body=# metrics unavailable: the OpenRegister AppHost observability engine is not installed
```

Content type is byte-identical to the engine's renderer, and the parent is now OCP:

```
CONTENT_TYPE=text/plain; version=0.0.4; charset=utf-8
matches_PrometheusRenderer=YES
parent=OCP\AppFramework\Controller
```

Auth posture read back off the compiled attributes — declared now, not inherited:

```
HealthController::index  attributes=[PublicPage,NoCSRFRequired]  required_present=YES  forbidden_absent=YES
MetricsController::index attributes=[NoCSRFRequired]             required_present=YES  forbidden_absent=YES (no NoAdminRequired, no PublicPage)
```

Metrics is therefore still admin-only by the deliberate absence of `#[NoAdminRequired]`, exactly as when it was inherited.

rubenvdlinde:
## Note: the `@spec` tags on these controllers were already dangling — carried forward verbatim, not fixed here

Both controllers carried `@spec openspec/changes/adopt-apphost/tasks.md#task-2.3`. That change was archived to `openspec/changes/archive/2026-06-16-adopt-apphost/`, so the path no longer resolves. I preserved the tags exactly rather than silently repointing them, because it is systemic rather than local:

| repo | distinct `@spec` targets under `lib/` | dangling |
|---|---|---|
| pipelinq | 124 | **70** |
| nldesign | 30 | **9** |

Repointing 70 files is not something to smuggle into a 500-every-route fix. Two observations for whoever picks it up:

- The canonical home exists in both repos (`openspec/specs/prometheus-metrics/spec.md`), which is where `@spec` is supposed to point — never at a change directory, precisely because change dirs get archived out from under the annotation.
- `phpcs` enforces that a `@spec` tag is *present*, not that its target *exists*. The gate that would catch this — hydra `spec-anchor-existence` — is `skipping` on this run (`enable-hydra-gates` is not set by the caller workflow), so nothing in CI sees it.

rubenvdlinde:
## `quality / E2E Tests (Playwright)` gave **no verdict** on this PR — and none on `development` either

Not a pass, not a failure. **Cancelled**, on both:

| run | branch @ sha | E2E started | E2E completed | elapsed | conclusion |
|---|---|---|---|---|---|
| 30882430080 | this PR @ `4b6e9a5` | 06:01:54Z | 06:47:12Z | **45m 18s** | `cancelled` |
| 30882429430 | `development` @ `15f2e4f` | 06:01:18Z | 06:46:34Z | **45m 16s** | `cancelled` |

Both hit the shared workflow's `timeout-minutes: 45` on the E2E job. A job killed by `timeout-minutes` is reported as **`cancelled`**, not `failure` — the log ends mid-run with Playwright still alive:

```
Terminate orphan process: pid (4796) (npm exec playwright test --config=playwright.config.ts)
Terminate orphan process: pid (24394) (chrome-headless-shell)
```

Two things follow:

1. **It is not caused by this PR.** The same job times out identically on plain `development` at the current head, and this change is PHP-only. It is red-on-arrival — or rather *grey* on arrival, which is worse, because a cancelled check reads as "not a failure" at a glance.
2. **pipelinq's E2E gate is currently producing no verdict on any run.** `quality / Quality Report` still reported **pass** on this run despite E2E being cancelled. Worth a separate issue: either the suite has to fit in 45 minutes or the cap has to move, but right now nobody is getting an E2E answer on any pipelinq PR.

### What I am relying on instead

Everything else on this run has its own conclusion read individually — **28 jobs, 0 failures**: 24 pass (including `PHPUnit` on all four legs, PHP 8.3/8.4 × NC stable31/stable32, and `phpcs`/`phpmd`/`psalm`/`phpstan`, all four of which pipelinq genuinely enables), 9 honestly skipped by job-level `if:`, and this one cancelled.

Plus the full `composer check:strict` locally in a PHP 8.4 container: `lint` 0, `phpcs` 0 errors, `phpmd` 0, `psalm` **No errors found!**, `phpstan` **No errors** (428 files), `phpunit` **1640 tests / 5179 assertions / 0 failures** — the last run *after* deleting `tests/Stubs/AppHost/`.

Merging on that basis, with the E2E no-verdict recorded here rather than glossed as green.

PR #685, opened 2026-08-04

rubenvdlinde:

This branch's caller job id was **not** `branch-protection`, so the shared branch-protection
workflow reported under the wrong context name, and the org ruleset requirement
`branch-protection / check-branch` never reported at all — not as a failure, as *nothing*.

**Several of these beta branches carry a second, independent fault**: the reusable workflow was
referenced as `Conduction/.github` instead of `ConductionNL/.github`. The `Conduction` org does not
exist (`GET /repos/Conduction/.github` -> 404), so on those branches the workflow could not even
resolve. Where present, that is fixed here too — renaming the job id alone would not have made the
context appear.

This matters for `beta -> main` promotion PRs, whose head is `beta`, so beta's tip is what
decides the context name and the workflow reference.

Does **not** change what is required and does **not** touch any ruleset or branch protection.

### DIFF
--- a/.github/workflows/branch-protection.yml
+++ b/.github/workflows/branch-protection.yml
@@ -5,5 +5,10 @@ on:
     branches: [main, beta]
 
 jobs:
-  check:
-    uses: Conduction/.github/.github/workflows/branch-protection.yml@main
+  # Job id must stay `branch-protection` so the check reports as
+  # `branch-protection / check-branch`, which is the context name the org
+  # ruleset requires. GitHub names a reusable-workflow context
+  # `<caller-job-id> / <called-job-name>`, so renaming this job silently
+  # detaches the required status check and leaves PRs permanently pending.
+  branch-protection:
+    uses: ConductionNL/.github/.github/workflows/branch-protection.yml@main


rubenvdlinde:
Closing as **superseded** — `beta` already contains this change, and 571 commits more.

**The proposed fix is already there.** `.github/workflows/branch-protection.yml` on `beta` declares the caller job id as `branch-protection:`, so the required context `branch-protection / check-branch` reports correctly. The two files differ only in that beta additionally has `permissions: {}` — beta is ahead, not behind.

**The second fault named in the description is also already fixed on beta.** All eight reusable-workflow callers (`code-quality`, `documentation`, `release-stable`, `release-beta`, `sync-to-beta`, `openspec-sync`, `issue-triage`, `branch-protection`) reference `ConductionNL/.github` there. Zero references to the non-existent `Conduction` org remain.

**Why this PR looked broken rather than redundant.** Its own `code-quality.yml` still called `Conduction/.github/...`, an org that 404s. An unresolvable reusable workflow does not fail the run — the job never starts, so the required `quality / Quality Report` context was simply *absent*, which on the PR page is indistinguishable from "still running". The merge box said BLOCKED and named a check that was never coming.

Fixing that on this branch made the quality workflow run for the first time, which then surfaced further problems (`appinfo/info.xml` failing schema validation, a vitest spec importing `src/services/recurringRevenue.js` which is absent *here* though present on beta, a stale `docs/features.json`). All of them are artefacts of this branch being three weeks stale — none reproduce on `beta`, whose `info.xml` validates cleanly against the appstore schema CI uses.

Nothing here needs to be carried forward. Closing rather than updating a 571-commit-stale branch to re-land a change that already shipped.

PR #686, opened 2026-08-04

rubenvdlinde:

`development` is the head branch of the `development -> beta` release PRs, so **development's
tip is what decides the context name on those PRs** — not beta's. With a caller job id other
than `branch-protection`, the org ruleset requirement `branch-protection / check-branch` never
reported on any release PR, leaving them permanently `BLOCKED`.

This renames the caller job id so the context matches. It does **not** change what is required and
does **not** touch any ruleset or branch protection.

### DIFF
--- a/.github/workflows/branch-protection.yml
+++ b/.github/workflows/branch-protection.yml
@@ -5,5 +5,10 @@ on:
     branches: [main, beta]
 
 jobs:
-  check:
+  # Job id must stay `branch-protection` so the check reports as
+  # `branch-protection / check-branch`, which is the context name the org
+  # ruleset requires. GitHub names a reusable-workflow context
+  # `<caller-job-id> / <called-job-name>`, so renaming this job silently
+  # detaches the required status check and leaves PRs permanently pending.
+  branch-protection:
     uses: ConductionNL/.github/.github/workflows/branch-protection.yml@main


rubenvdlinde:
Merging with `--admin`. Recording why, since `E2E Tests (Playwright)` is red.

**This PR changes exactly one file — `.github/workflows/branch-protection.yml` — renaming the caller job id from `check` to `branch-protection`.** It cannot affect Playwright results.

**E2E has never produced a passing verdict on `development`.** The last four Code Quality runs all ended `cancelled` (05:59, 05:59, 16:41, 13:05) — hitting the 45-minute cap, which is *no verdict at all*. This run finally produced a real `failure` instead, which is strictly more information than a cancellation.

The underlying hang is tracked in **#675**, which parallelises the suite (measured 56.9 min → 17.8 min with `fullyParallel: true` + 6 workers, after verifying spec isolation — 3 of 44 specs do real CRUD, so this was checked rather than assumed).

**25 pass · 8 skipped · 1 fail.** The one failure is pre-existing and unrelated to a workflow-file rename.

### Why this change matters

pipelinq's `main`/`beta` are covered by an org ruleset requiring the context **`branch-protection / check-branch`**. The caller job was named `check`, so it emitted **`check / check-branch`** — a name that can never match. The required context was therefore *absent*, not failing, leaving every PR permanently `BLOCKED` and shippable only by `--admin`.

That defect spans ~16 repos. The `main` and `beta` counterparts of this fix (#684, #685) are **deliberately left for a human** — they target protected branches, where the review gate genuinely applies.

Issue #688, opened 2026-08-04
E2E spec debt: 14 tests assert surfaces retired by the 2026-07 IA revision


rubenvdlinde:
## Summary

Fourteen e2e tests still assert navigation entries and pages that were **deliberately retired** by the 2026-07 IA revision (`src/menu-layout.json`) and the `unify-ticket-supertype` consolidation. They fail because the product changed, not because it is broken.

They were invisible until #675 made the shared Playwright job produce a verdict at all — before that the job hit `timeout-minutes: 45` and reported `cancelled`, which is no verdict.

Each has a documented replacement, recorded below so the rewrite is mechanical. I did not patch them in #675 because I could not verify the replacement DOM contracts in that session, and a guessed locator is worse than a known-red test.

## 1. Retired features — nav entry AND pages removed

`src/menu-layout.json#_removalsNote` is the authority.

| Spec | Tests | Documented replacement |
|---|---|---|
| `spec-coverage/barcode-lookup.spec.ts` | 2 | *"'Barcode lookup' is dropped in favour of searching the Products index by barcode, which is one surface instead of two for the same question."* -> assert barcode search on the Products index. |
| `spec-coverage/billing-categories.spec.ts` | 3 | *"'Reports & Compliance' (AnalyticsGroup) and its sole child 'Billing categories' are dropped ... The billingCategory SCHEMA and its objects stay (ShillinqWipService and the 'Hours by billing category' widget on the Operational dashboard still read them); only the nav entry and its three pages go."* -> assert the **"Hours by billing category" widget on the Operational dashboard**. |

Verified absent: no `menu` entry and no `pages` entry for either, across `src/manifest.json` and every `src/manifest.d/*.json`.

## 2. Moved to the Nextcloud admin page

`src/menu-layout.json#_adminSettingsNote`: *"Genuinely-admin configuration no longer lives in the app nav at all ... Moved there: Messaging ..., **CTI telephony settings**, Payment providers (PSP), POS tender types, POS medewerkers, POS rollen. Their in-app pages and nav entries are gone; the Vue components are mounted as sections on the admin page instead."*

| Spec | Tests | Replacement |
|---|---|---|
| `spec-coverage/cti-integration.spec.ts` | 4 | Assert the CTI sections on `/settings/admin/pipelinq` (mount host `#pipelinq-settings`, see `templates/settings/admin.php`) instead of a sidebar entry. |

## 3. Consolidated into the unified Tickets index

[...]

Issue #687, opened 2026-08-04
Channel analytics and Agent performance pages are unreachable from the UI


rubenvdlinde:
## Summary

Two fully-built reporting pages have **no entry point anywhere in the UI**. They render correctly, but a user can only reach them by typing a hash URL by hand.

| Page | Route | Manifest id |
|---|---|---|
| Channel analytics | `/rapportage/channels` | `ChannelAnalyticsView` |
| Agent performance | `/rapportage/agents` | `AgentPerformanceView` |

Both are complete `type: "dashboard"` pages in `src/manifest.json` with dedicated body-widget components (`src/components/rapportage/ChannelComparisonSection.vue`, `AgentPerformanceSection.vue`), each self-fetching `/api/rapportage/channels` and `/api/rapportage/agents`.

## Evidence

Checked all three ways a pipelinq page can be reached — none of them expose these two:

1. **No menu entry.** Neither `ChannelAnalyticsView` nor `AgentPerformanceView` appears as a `route` on any menu entry across `src/manifest.json` or any `src/manifest.d/*.json` fragment. The only reporting menu entry is `Rapportage` (label "Reporting", route `Rapportage`).
2. **No in-page link.** `grep -rn "rapportage/channels\|rapportage/agents" src/` matches only `_note` prose and the components' own axios calls — no router-link, no navigate action, no button.
3. **No deep link.** `manifest.deepLinks` registers only `client`, `lead`, `ticket` and `contact`.

## How it surfaced

Two e2e tests in `tests/e2e/rapportage.spec.ts` fail against this:

- `rapportage page navigates to channel analytics`
- `rapportage page navigates to agent performance`

Both wait for `.rapportage-links` — the in-page button strip that the old bespoke `RapportageDashboard.vue` rendered. That component was replaced by the declarative dashboard (change `pipelinq-dashboards-declarative`) and the link strip was dropped along with it, but no replacement affordance was added.

Note the two sibling tests `channel analytics page loads` and `agent performance page loads` **pass** — they deep-link the hash directly. So the pages work; only the way in is missing. That is why this reads as a test failure rather than an obviously broken page.

## Why this needs a product decision (not a test patch)

The failing assertions are correct: they assert that a user can navigate to these reports. Rewriting them to deep-link the hash would make CI green while leaving the reports unreachable for real users, so they have deliberately been left failing.

Options:
[...]

PR #693, opened 2026-08-04
fix(e2e): first-visit walkthrough blocked clicks; settings-foldout nav matched @nextcloud/vue 8 markup

rubenvdlinde:
## Baseline

Run `30898348537`, job `91956896164` (`E2E Tests (Playwright)`), **20 m 18 s** against a 45-minute cap — a real verdict, not a timeout. **39 failed / 132 passed / 40 skipped / 1 did not run.**

## Clustering the 39 first

Grouping the 46 `error-context.md` attachments by the *identical page snapshot* they captured collapses the 39 failures into far fewer causes than tests. Three of those clusters are fixed here; the rest are itemised below and left red on purpose.

**1 — the first-visit product tour blocks clicks (2 tests).**
`src/manifest.json#walkthrough` declares `pipelinq:getting-started` with `trigger: "first-visit"`. On a CI instance nobody has ever completed it, so it starts on the first page load and `CnWalkthrough` paints `div.cn-walkthrough__dim--full` across the viewport. Playwright's log is unambiguous:

```
- element is visible, enabled and stable
- <div class="cn-walkthrough__dim cn-walkthrough__dim--full"> from
  <div role="dialog" aria-modal="true" class="cn-walkthrough" aria-label="Welcome to Pipelinq">
  subtree intercepts pointer events
- retrying click action   (153 ×)
```

The target is never *unclickable* in a way Playwright fails fast on — it retries for the whole test timeout. `spec-coverage/products.spec.ts:34` burned 60 s, `workflows/product-crud.spec.ts:112` 90 s.

`ci-seed.sh` already prints `first-time setup OK — the shell will render and no wizard will cover it`, and that is true: the **setup wizard** is completed there. The **walkthrough** is a different mechanism with its own `completionConfigKey`, and nothing ever completed it.

Fixed by ending the tour through its real affordance — the `.cn-walkthrough__close` button, aria-label "Close tour", whose handler marks the tour complete so it does not reopen. Preferred over Escape/backdrop, which only dismiss the current showing.

**2 — the settings-foldout fallback was written against @nextcloud/vue 8 (3 tests).**
`menu-layout.json#settingsSection` promotes `Pipelines` into `NcAppNavigationSettings`. `revealNavEntry()` handled that with

```js
page.locator('#app-navigation-vue button.settings-button[aria-expanded="false"]')
```

`settings-button` is 8.x markup. pipelinq resolves **9.9.0** (`package.json ^9.9.0`, `package-lock.json` 9.9.0), and in 9.x `NcAppNavigationSettings` renders an `NcButton` with CSS-module classes — the built chunk carries `_button_ypW42`, and `settings-button` appears **nowhere** in `dist/`. `.isVisible()` was false, the branch never ran, the leaf stayed hidden, and all three `pipelines.spec.ts` tests failed with

```
locator resolved to <a title="Pipelines" href="#/pipelines" …>  - unexpected value "hidden"
```

reported against the Pipelines *page*. Same shape as the group-caption bug already documented in this file: **the navigation helper failed and the leaf took the blame.** Now matched on our own `data-testid="cn-nav-settings"` + `aria-expanded`, with the 8.x class kept as a second alternative.

**3 — an assertion that no render could satisfy (1 test).**
`returns.spec.ts:36` asserted `getByRole('radio', { name: 'Table' })`. `CnActionsBar.vue` (lines 22-55) renders the view switch as `<div role="group" aria-label="View mode">` containing plain `<button type="button" :aria-pressed>` segments. There is no `role="radio"` in the component, so that assertion failed against the *correct* page. Rewritten to the role the component actually exposes; the claim is unchanged.

## What this PR does NOT do

- **#687** (2 tests) and **#688** (14 tests) stay red. Untouched.
- No `.skip`, no `retries`, no raised timeout, no widened allow-list, no deleted test, no suppression, no baseline. The `showAdd`, heading and seed clusters below are **filed, not patched** — see the linked issues.
- No application code is changed; the diff is two test files.

## Still red, with issues filed

| cluster | tests | issue |
|---|---|---|
| POS/Returns assert a primary CTA that `showAdd: false` deliberately removed | 5 | filed |
| index pages render no accessible heading (`CnIndexPage` `showTitle` defaults false) | 4 | filed |
| Contacts/Tasks have no seeded objects; `queues` asserts a `Next` button over 3 rows | 3 | filed |
| dashboard widgets renamed/removed (`Customer Satisfaction`, `Recurring revenue (MRR)`) | 2 | filed |
| `client-crud` sidebar search input stays hidden | 1 | filed |

## Verification

Counts and job ids for the run on this branch are reported in the thread. Baseline for every "pre-existing" claim above: run `30898348537`, job `91956896164`, on `development`.


rubenvdlinde:
## Merged — with a mutation test proving the fixed specs can still fail

A repaired e2e test is worth nothing if it cannot go red, so before merging I re-broke the code under test and confirmed the specs fail again.

### Mutation test

I cut `test/mutation-e2e-foldout-selector` from this PR's head and reverted **one variable only** — `openSettingsFoldout()` back to the `@nextcloud/vue` 8-only selector this PR replaced:

```js
// mutation: the v8-only selector
page.locator('#app-navigation-vue button.settings-button[aria-expanded="false"]')
```

Everything else (the walkthrough dismissal, the `returns.spec.ts` role fix) was left in place, then dispatched through the normal `Code Quality` workflow.

| spec | `development` (no fixes) | this PR | mutation (foldout reverted only) |
|---|---|---|---|
| `pipelines.spec.ts:13` | **FAIL** | pass | **FAIL** ← went red |
| `pipelines.spec.ts:35` | **FAIL** | pass | **FAIL** ← went red |
| `products.spec.ts:34` | **FAIL** | pass | pass |
| `returns.spec.ts:36` | **FAIL** | pass | pass |

The two Pipelines specs went red exactly as predicted, which is the direct proof for the foldout fix. The other two **stayed green** while the foldout was reverted, which attributes them cleanly to the other two changes — `products.spec.ts:34` to `dismissWalkthrough()` and `returns.spec.ts:36` to the role correction. One run, single variable, full attribution. The mutation branch has been deleted.

### Suite numbers, identically conditioned

| ref | failed | passed | skipped |
|---|---:|---:|---:|
| `development` @ `947fd4d1` | 35 | 136 | 40 |
| `development` @ `603d412d` | 35 | 136 | 40 |
| mutation branch | 33 | 137 | 40 |
| **this PR @ `5496bd7`** | **31** | **140** | 40 |

Four specs fixed, **zero new failures**. The suite is 31-red because `development` is 35-red — a pre-existing debt this PR reduces but does not clear.

### On the `returns.spec.ts` assertion change

Swapping `getByRole('radio', { name: 'Table' })` for `getByRole('button', …)` is a **correction, not a weakening**. I verified against the component source: `CnActionsBar.vue` renders the view switch as `<div role="group">` containing `<button :aria-pressed>` segments, and there is **no `role="radio"` anywhere in `@conduction/nextcloud-vue`**. The original assertion was structurally impossible — it could never have matched any render, correct or broken. The new one asserts the same claim against the contract the component actually implements.

### CI state at merge (`5496bd7`)

- **Hydra Gates: SUCCESS** — `ALL 58 APPLICABLE GATES PASSED — and all 58 of them ran`, with 4/6/7/24/33 reported NOT APPLICABLE. This job's earlier failure was never a gate defect: on the older gates version those reads came back as "did not run", which `hydra-gates-require-full-coverage` treats as failure. Rebasing onto current `development` (v1.3.0, which separates NOT APPLICABLE from SKIPPED) cleared it with no change to this PR's content.
- 24 success / 5 skipped / 2 failure. The two failures are `E2E Tests (Playwright)` (pre-existing debt, reduced by this PR) and `Quality Report`, which is a pure aggregator of the former.

Merged with `--admin --squash` under the standing authorisation for `development`, since the only red is pre-existing and the diff is test-only.

Issue #700, opened 2026-08-04
Suppression & baseline audit: what the green PHPMD number hides


rubenvdlinde:
Audit-only measurement of every suppression and baseline in this repo. **Nothing was removed or changed** — all numbers come from throwaway copies re-run in a container. Part of a 22-repo fleet sweep; see the summary issue for the fleet picture.

## PHPMD: reported vs true

| arm | violations | what it means |
|---|---|---|
| **reported** (`composer phpmd` as configured) | **0** | what CI prints today |
| baseline neutralised | 44 | + what `phpmd.baseline.xml` hides |
| baseline + `@SuppressWarnings` neutralised | **456** | true current state |

**Delta: 0 → 456 (+456).**

`phpmd.baseline.xml` holds **39** entries, wired via `--baseline-file`. Note it stays active even if that flag is dropped — PHPMD auto-discovers this filename.

## Baselines

| file | entries | wiring |
|---|---|---|
| `phpmd.baseline.xml` | 39 | flagged |
| `psalm-baseline.xml` | 0 | n/a |
| `phpstan-baseline.neon` | 39 | included |

## Inline suppressions (`lib/`)

- `@SuppressWarnings`: **428** — 369 carry a written reason, **59 are bare**.

Triage on whether a reason exists **and still holds** — not on the rule name.

## ⚠️ Inert `<exclude-pattern>` (config bug)

`phpmd.xml` nests 1 `<exclude-pattern>` **inside** a `<rule>` element. **PHPMD only honours `<exclude-pattern>` as a direct child of `<ruleset>`**, so this exemption has never applied.

Shared-template defect across 19 of 21 repos — fix once in the template.

## Suggested burn-down order

1. **Fix the exclude-pattern** (above) — cheapest; retires interface-mandated noise legitimately.
2. **Triage the 59 bare `@SuppressWarnings`** — each gets a reason or goes.
3. **Resolve orphan/undeclared baselines** — they mislead without protecting.
4. Burn down the remainder by rule cluster, largest first.

### Caveats measured, not assumed

- `UndefinedVariable` here is overwhelmingly **false positives** from by-ref out-params (`preg_match($re,$s,$matches)`, `openssl_*(output: $var)`). The rule name is not a defect signal.
- Much `UnusedFormalParameter` mass is **interface-mandated** (Migration steps, listener signatures) and cannot be removed.

---
[...]

PR #704, opened 2026-08-05
chore(quality): burn 14 PHPMD findings, scope the UnusedFormalParameter leg, shrink the baseline 39→27

rubenvdlinde:
## PHPMD: reported 0, actually 44 — now 30, with the baseline honest about it

Part of the fleet suppression audit (ConductionNL/.github#155). Audit issue: #700.

`phpmd.baseline.xml` made `composer phpmd` — and the CI PHPMD job — report a green zero over **44 live findings**. This PR fixes 14 of them, replaces three opaque baseline entries with one scoped and documented ruleset exclusion, and **shrinks the baseline from 39 entries to 27** so the gate genuinely covers the files that are now clean.

The baseline is **RETAINED**, not deleted — see "What remains" below.

### Numbers

PHPMD 2.15.0 / PHP 8.4.22, exit code read directly (`0` clean, `2` violations, `1`/`255` tool failure):

| | findings | exit |
|---|---|---|
| reported today (baseline present) | 0 | 0 |
| **baseline deleted, before this PR** | **44** | 2 |
| **baseline deleted, after this PR** | **30** | 2 |

### Triage of the 44

| bucket | count | detail |
|---|---|---|
| **interface-mandated** | 3 | `run($argument)` on three `TimedJob` subclasses. `OCP\BackgroundJob\TimedJob` declares `abstract protected function run($argument)`; a job needing no argument still cannot drop the parameter. |
| **genuine debt (fixed here)** | 14 | 7 MissingImport, 5 ElseExpression, 2 LongVariable |
| **genuine debt (remaining)** | 27 | 6 WeightedMethodCount, 8 CyclomaticComplexity/NPath pairs, 4 LongMethod, 2 LongParameterList, 2 StaticAccess, 1 TooManyPublicMethods, 1 CouplingBetweenObjects |
| **structurally unreachable** | 0 | — |

### What changed

**MissingImport ×7** — `use RuntimeException;`, `use DateTime;`, `use DateTimeImmutable;`, `use DateInterval;` added and the leading backslash dropped. All are global PHP classes, so no new autoload edge is introduced.

**ElseExpression ×5** — four become initialise-then-override (`ComplaintSlaService::calculateDeadline`, `EmailSyncService::setSyncEnabled`, `ReportingService::getChannelTrend` and `::calculateSlaCompliance`). No ternaries — `Squiz.PHP.DisallowInlineIf` forbids them.

`DemoSeedService::seed` deliberately does **not** use that shape. Both of its branches **write** (`ObjectService::saveObject` / `TicketService::save`), so initialise-then-override would have executed both and **double-written every seeded object**. It is extracted to `saveSeedObject()` instead, which performs exactly one write per call.

**LongVariable ×2** — `$identityRegistryClass` / `$regulatorRegistryClass` → `$identityClass` / `$regulatorClass` in `Application::wireGdprSeamProviders()`.

**`phpmd-unusedparams.xml` (new)** — `UnusedFormalParameter` moves into its own ruleset with a **top-level** `*/BackgroundJob/*` exclude-pattern, and `composer phpmd` gains a second leg (worst exit code wins, neither leg can short-circuit the other).

This matters because the pattern that used to sit **nested inside the `<rule>`** in `phpmd.xml` was **inert**: PHPMD honours `<exclude-pattern>` only as a direct child of `<ruleset>`. It also pointed at `*Migration*`, and this repo has no `lib/Migration` at all — dead config twice over.

Positive-controlled both ways:
- a `$unusedParam` in `lib/Service/` **is** reported by the second leg (exit 2)
- the same in `lib/BackgroundJob/` is **not** — the exclusion is scoped as intended
- an `ElseExpression` + `MissingImport` + `UnusedLocalVariable` planted in `lib/BackgroundJob/` **is** reported by the main leg, so isolating the rule did **not** blind every other rule to that directory

### The baseline: retained, but shrunk and documented

Twelve entries are removed — nine because the findings are **fixed**, three because the exclusion moved to the scoped ruleset. This is not cosmetic: **a baseline entry with no `method` attribute suppresses every finding of that rule in that file, including future ones.** A stale file-scoped entry is a dead gate, not a harmless leftover.

The file keeps a header recording the measured numbers and exactly what remains. Positive-controlled: a fresh violation planted in a file whose entry was removed is reported and exits 2, so the shrunken baseline is a real ratchet.

### What remains, and why the file stays

**30 findings.** The blocker is the six `WeightedMethodCount` classes (`Application` 66, `SemanticHandoffController` 50, `MigrateAvgVerzoekenToOrDsar` 81, `DemoSeedService` 62, `TimeBillingHandoffService` 82, `TrackingLinkService` 79, threshold 50). WMC is the **sum of every method's cyclomatic complexity**, so splitting a method *in place* makes it **worse** — only moving logic to another class helps. Each needs a cohesive-collaborator extraction with its own review; that is a behaviour-bearing refactor, not a suppression sweep, and is deliberately out of scope here.

Deleting the baseline today would redden `development` for everyone with 30 real findings. Per the audit's rule, an honest "44 → 30, here is what remains and why" beats a forced green or a shipped red.

### Tests

| | before | after |
|---|---|---|
| tests | 1640 | 1640 |
| assertions | 5179 | 5179 |
| warnings / skipped | 3 / 13 | 3 / 13 |

`phpcs` 0 errors (one pre-existing alignment error in `Application.php` fixed on the way), `psalm` 0, `phpstan` 0.

### What I did NOT do

- did not delete `phpmd.baseline.xml` — 30 real findings remain
- did not add a `@SuppressWarnings`, a baseline entry, a threshold change, a `.skip` or a weakened assertion
- did not touch any rule threshold in `phpmd.xml`
- did not attempt the six `WeightedMethodCount` class extractions
- did not change behaviour anywhere

### Pre-existing CI state

`development` is already red: **E2E Tests (Playwright)**, **Hydra Gates** and the **Quality Report** aggregate all fail on the last completed Code Quality run (31004087489). Those are not caused by this PR. (The run before it is `cancelled`, which is no verdict.)


rubenvdlinde:
**Dedup survivor — but held, blocked on gates its own diff pulled into scope.**

This PR is the survivor of the duplicate pair (see the closing comment on the sibling PR); its exclude-pattern was tightened to the fleet-canonical `*/lib/…/*` form before this run, positive-controlled with PHPMD 2.15 / PHP 8.4:

| probe | `*/lib/X/*` | `*/X/*` |
|---|---|---|
| unused param in `lib/Service/` (control) | **reported** | **reported** |
| unused param in the app's own `lib/X/` | not reported (intended) | not reported (intended) |
| unused param in `lib/Service/X/` | **reported** | **silently swallowed** |

I also confirmed the PHPMD baseline is honoured on the second leg **without** an explicit `--baseline-file`, so the two-leg script is baseline-neutral. Controlled three ways — flag absent: exit 0; flag present: exit 0; **baseline file removed: 3 findings, exit 2**. The control proves the arms can differ, so the first two zeros mean something.

**Why it is not merged:** its own Hydra Gates run fails on gates that its diff pulled into scope:

```
[gate-46] spec-anchor-existence: FAIL — unresolved @spec target(s)
[gate-47] security-change-has-tests: FAIL — security-touching change(s) without a test co-change
```

The gates are diff-scoped per ADR-020, so touching a file to add a `use` statement or drop a `@SuppressWarnings` legitimately brings that file's pre-existing `@spec` debt into scope. That is the intended "you touched it, you own it" behaviour, and I am not going to add a waiver, a baseline entry, or weaken `require-full-coverage` to get around it.

I did **not** cite `development` as "also red" here. A push to the base branch scopes 0 files and Hydra Gates passes in ~20s having inspected nothing — a vacuous green that proves nothing. Only a genuinely red base is evidence, and this failure is scoped to files only this PR touched, so no base run could speak to it either way.

**gate-47 is worth a second look by the gate owner.** This PR's `lib/` changes are `use` statements and removed suppression tags — no behaviour change — yet gate-47 asks for a test co-change. The same false positive is blocking openregister#2355, where the diff is provably **comment-only** (all 30 changed `lib/` lines are docblock). A gate that demands tests for an import or a comment is asking for a test of nothing.

**To land this:** repoint the unresolved `@spec` anchors onto real headings (same technique as openregister#2355), and get a ruling on gate-47 for non-behavioural diffs. The PHPMD work itself is measured and sound.

Issue #160, opened 2026-03-25
[klachtenregistratie] 1.2 Create ComplaintSlaJob background job


rubenvdlinde:
**Spec ref**: REQ-KL-010
**Tracking**: #150

Create `lib/BackgroundJob/ComplaintSlaJob.php`:
- Extends `TimedJob`, runs every 15 minutes
- Queries open complaints, checks SLA deadlines via ComplaintSlaService
- Logs overdue complaints via LoggerInterface

Register in `appinfo/info.xml` background-jobs section.

**Acceptance**: Job runs periodically, logs overdue complaints.

PR #706, opened 2026-08-05
docs(security): stop two orphaned guards from claiming enforcement they do not provide

rubenvdlinde:
## Summary

pipelinq has three gate-6 (`orphan-auth`) findings. This PR **fixes none of them by wiring**, deliberately, and instead removes a docblock that claims enforcement which does not exist. Gate-6 still reports all three afterwards — nothing here is a suppression.

## `ZgwCoexistenceValidator`

Its class docblock said: *"Called before any createZaak flow ... so the pipeline blocks before either external call is issued."* Neither half is true at HEAD.

1. `validateWritePath()` has **zero callers**. So does `ZrcClient::createZaak()` — the "createZaak flow" it claims to protect is itself unwired, so there is no call site to insert the check into.
2. **It could not fire even if it were called.** The `stufEndpoint` schema was relocated to procest by the archived `2026-06-23-pipelinq-stuf-zkn-removal` change. `activeStufWriters()` looks that schema up, always gets `[]`, and the `$zgwWriters !== [] && $stufWriters !== []` branch is therefore unreachable.

Point 2 is why this is **not** force-wired. A guard that is structurally incapable of rejecting anything would turn the gate green while protecting nothing — the exact defect the orphan-auth programme exists to remove, and a direct violation of "a passing assertion is evidence about the assertion until you have shown it can fail."

The docblock now records what restoring real enforcement would take, and corrects one point from the archived `2026-07-16-orphan-auth-remediation` design: that change reasoned there was *"no pipelinq-owned save hook to wire it into"* because ZGW endpoints are OpenRegister-object-managed. There **is** a hook — OpenRegister's pre-persist, stoppable `ObjectCreatingEvent`/`ObjectUpdatingEvent` pair, the same mechanism `LocationBagValidationListener` uses in procest. It is unusable here only because of point 2, not because the hook is missing.

## `PortalTenantService::isSelfSignupAllowed`

Reads a live tenant schema field (`register.d/40-portal.json`) that nothing consumes. The portal has no self-signup endpoint at all: `PortalAccountService` only closes accounts, `PortalAuthController` only authenticates existing ones, and no controller or service creates a `portalAccount`.

The docblock now says so, and says explicitly that the remedy is to build the flow and call this from it — **never** to invent an account-creation endpoint in a citizen portal in order to give a predicate a caller. Inventing that endpoint would create an unreviewed account-creation surface: strictly worse than the finding.

## Verification

- Absence claims positive-controlled: the same search that finds **zero** self-signup / `createZaak` call sites finds `isWidgetOriginAllowed` in `PortalRequestGuard` and its tests, so the search is not silently matching nothing.
- Gate-6 before and after: **3 findings, unchanged**. Doc-only diff, no behaviour change, no `@SuppressWarnings`, no baseline entry, no waiver.

Fleet context and the full 19-repo re-measurement are in ConductionNL/.github#158.


rubenvdlinde:
### Status — leaving this open for a human, deliberately

Run 31015733932 has **31 jobs**: 28 green, and three red — `quality / E2E Tests (Playwright)`, `quality / Hydra Gates`, `quality / Quality Report`.

**All three are pre-existing.** The last pre-outage run on `development` (31004087489, 2026-08-05T12:04:51Z) has the same 31 jobs and **the same three failures**, byte for byte. Both runs are outside the `.github` d68fb727 window (13:29:41Z–13:47:44Z) that produced zero-job non-verdicts, and both are named `Code Quality` rather than the raw workflow path, so neither is one of those.

The diff here is **PHPDoc only** — two docblocks, no executable line changed. It cannot cause a Playwright failure, and it cannot change what a static gate finds.

Two reasons this is not being self-merged:

1. `quality / Hydra Gates` is a gate I am **actively disputing** in ConductionNL/.github#158 and #160 — including the pipelinq gate-6 findings this very PR declines to fake-wire. Merging past a gate on the strength of my own argument that the gate is wrong is not a call to make unilaterally.
2. `development` is red on all three, so there is no green baseline to merge into, and nothing here improves that.

Confirmed locally against the pushed tree: gate-6 still reports all three pipelinq findings, unchanged — this PR adds no suppression, no baseline entry, no threshold change. That is intentional. The code now simply stops *claiming* enforcement it does not perform.

Merge whenever a maintainer is happy with the reasoning; there is no CI signal here that this PR needs to wait for.

PR #711, opened 2026-08-05
chore(apphost): register OpenRegister's autoloader before touching AppHost

rubenvdlinde:
## What

Adds a short prelude as the first statement of `register()` in `lib/AppInfo/Application.php` that registers OpenRegister's PSR-4 prefix with the autoloader before any AppHost `class_exists()` probe runs.

## Why — the load-order defect

`OC_App::getEnabledApps()` does `sort($apps)`. `Coordinator::registerApps()` walks that **sorted** list, calling `OC_App::registerAutoloading($appId, $path)` and then `$application->register()` **one app at a time**.

So each app's `register()` runs *before* the PSR-4 prefix of every alphabetically-**later** app exists. For any app sorting before `openregister`, `class_exists('OCA\OpenRegister\AppHost\...')` answers **FALSE** at `register()` time.

The failure mode is nasty: an unguarded reference throws an `\Error` that aborts the **entire** `register()`. `Coordinator` catches it, logs an `emergency`, and continues — so the app stays enabled and looks perfectly healthy while half its wiring silently never registered.

## Honest scoping — this app is safe today

**This app sorts *after* `openregister` alphabetically, so the `class_exists()` probes already answer TRUE. It is safe today — by alphabet alone.**

This PR is **landmine removal, not a fix for an observed failure in this repo**:

- It is a **no-op** against today's app id.
- It stops the app depending on a property of its own name. A future rename, or a fork/whitelabel with an id sorting before `openregister`, would silently arm the defect.

The defect *was* observed live in `doriath`, and independently in `openconnector` — both of which sort *before* `openregister`.

## Why this shape

- `registerAutoloading()` touches **only** the autoloader and is idempotent (guarded by the `$alreadyRegistered` key), so calling it here is cheap and safe to repeat.
- Deliberately **not** `IAppManager::loadApp()`, which would mark OpenRegister loaded and boot it *before its own `register()` had run*.
- Wrapped in `try { … } catch (\Throwable)` so an absent/disabled OpenRegister falls through to the existing degraded path unchanged.

## Scope

- Exactly one file, +17 lines, nothing removed. No existing `class_exists()` guard or comment was touched, and nothing was reformatted.
- `appinfo/routes.php` deliberately **not** touched — route files are loaded by the router long after every app has registered, so they are safe by timing.

## Verification

- `php -l` clean.
- PHPCS run with the container's PHP 8.4 against this repo's own `phpcs.xml` (incl. the custom `NamedParameters` / `NoLegacyServerAccessors` / `SpecTag` sniffs), on the unmodified file **and** the modified file: **identical results — this change introduces zero new findings.**


rubenvdlinde:
Closing **for pipelinq only**. This verdict is repo-scoped and says nothing about the other five repos in this sweep — in several of them the prelude is genuinely load-bearing. Reopen if you disagree; the measurements are below so they can be checked rather than taken on trust.

Two independent reasons, both measured against `origin/development` at `47a1a590f`.

**1. The hazard this guards against does not exist in pipelinq's code.**

The added comment says *"the AppHost `class_exists()` probes below answer TRUE today — by alphabet alone"*. There are no such probes:

```
$ grep -rn "class_exists" lib/AppInfo/
(no matches)
```

pipelinq's only AppHost reference is a **string inside a lazy DI service closure** (`Application.php:364`, `registerAppHost()`), resolved at container-resolution time, long after every app has registered. `check_apphost_autoload_prelude.py` explicitly excludes exactly this shape: *"Lazy service closures that merely mention an AppHost class are NOT flagged."* The one `AppHost\Bootstrap` mention in the file is in a docblock (line 318).

Accordingly **gate-64 already passes on `development` without this change** — and it passes for the right reason, not because it is blind. Positive control, on this same tree:

```
gate-64 on development, unmodified                              -> OK,   exit 0
gate-64 after inserting an eager class_exists('...AppHost\Bootstrap')
                                                                -> FAIL, exit 1
   "LATENT: 'pipelinq' sorts AFTER 'openregister', so this happens to
    work today — by alphabet alone... Add the ADR-040 prelude"
```

So the checker can fail on this tree; it simply has nothing to flag here.

**2. The change introduces a phpmd failure that `development` does not have.**

`quality / PHP Quality (phpmd)` fails on this PR (job 92870429394, step `Run phpmd`) while it passes on `development`. Reproduced locally, and — importantly — proven *not* pre-existing by running the checker against the base tree with the same file forced into scope, rather than by comparing counts:

```
development, lib/AppInfo/Application.php, phpmd.xml + baseline   -> EXIT=0 (clean)
this change applied onto CURRENT development, same command       -> EXIT=2
  lib/AppInfo/Application.php:134  StaticAccess
  Avoid using static access to class '\OC_App' in method 'register'.
```

(Checked on current `development` specifically: this branch is 2 commits behind and predates `phpmd-unusedparams.xml`, so the PR's own CI ran a different phpmd config than the one it would meet on merge. The `StaticAccess` finding appears under both.)

There is no clean way to keep the change and the gate: the only silencing move is a `@SuppressWarnings(PHPMD.StaticAccess)` waiver, and a gate made to pass by making it blind is worse than the finding.

Net: for pipelinq this trades a real, reproducible phpmd regression for a guard against a hazard the code does not have.

PR #728, opened 2026-08-08
fix(security): declare the auth posture on 17 admin-only routed endpoints

rubenvdlinde:
fix(security): declare the auth posture on 17 admin-only routed endpoints

hydra gate-5 (route-auth), measured FULL-TREE at origin/development, reported
17 routed methods with no auth attribute. Every one is admin-intended, but the
posture was declared only by OMISSION — Nextcloud's SecurityMiddleware default.
Nothing in the source stated it, so the intent was unreviewable and one added
attribute would silently have widened access.

  BerichtenboxAdminController  retry, stats
  PortalAdminController        saveConfig, accounts, auditEvents
  PosPaymentController         index, show, update, test
  PosRoleController            create, update, destroy
  PosStaffController           index, show, create, update, destroy

All 17 now carry #[AuthorizedAdminSetting(Application::APP_ID)] — already this
repo's dominant convention for admin endpoints (39 pre-existing usages;
MessagingController declares its provider config/test endpoints exactly this
way, which is the same shape as PosPaymentController's).

WHY THIS DOES NOT WIDEN ACCESS. SecurityMiddleware (lib/private/AppFramework/
Middleware/Security/SecurityMiddleware.php:137-160) authorizes an
AuthorizedAdminSetting method when isAdminUser() is true, or when the user has
delegated authorization for one of the named SETTING CLASSES, looked up via
groupAuthorizationMapper->findAllClassesForUser(). The argument here is the app
id 'pipelinq', which is not a settings class and can never appear in that list,
so the delegated branch cannot succeed. Net posture is admin-only — identical
to the previous framework default — plus NC's admin-IP restriction
(remoteAddress->allowsAdminActions()), which is strictly tighter than before.

The 12 POS methods additionally keep their in-body requireAdmin() guard, so
they are gated twice. The 5 methods in the two Admin controllers have no
in-body guard, which is exactly why APP_ID (no delegation possible) was used
rather than a real Settings class.

PROOF THE GATE CAN FAIL, not just that it passes:
  before          gate-5 FAIL, 17 findings (the list above)
  after           gate-5 PASS
  after + mutate  removed the attribute from PosStaffController::destroy only
                  -> gate-5 FAIL, 1 finding:
                     lib/Controller/PosStaffController.php:190 method=destroy
                     rule=missing-auth-attribute
                  attribute restored.

gate-9 (semantic-auth) stayed at 7 findings across the change, so no new
attribute-vs-body mismatch was introduced. Full-tree gate count 19 -> 18.

### DIFF
--- a/lib/Controller/BerichtenboxAdminController.php
+++ b/lib/Controller/BerichtenboxAdminController.php
@@ -35,7 +35,9 @@ use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\Attribute\AuthorizedAdminSetting;
 use OCP\AppFramework\Http\JSONResponse;
 use OCP\IAppConfig;
+use OCP\IGroupManager;
 use OCP\IRequest;
+use OCP\IUserSession;
 use Psr\Container\ContainerInterface;
 use Throwable;
 
@@ -47,17 +49,42 @@ class BerichtenboxAdminController extends Controller
     /**
      * Constructor.
      *
-     * @param IRequest           $request   Request.
-     * @param ContainerInterface $container DI container.
-     * @param IAppConfig         $appConfig App config.
+     * @param IRequest           $request      Request.
+     * @param ContainerInterface $container    DI container.
+     * @param IAppConfig         $appConfig    App config.
+     * @param IUserSession       $userSession  The user session.
+     * @param IGroupManager      $groupManager The group manager.
      */
     public function __construct(
         IRequest $request,
         private readonly ContainerInterface $container,
         private readonly IAppConfig $appConfig,
+        private readonly IUserSession $userSession,
+        private readonly IGroupManager $groupManager,
     ) {
         parent::__construct(appName: Application::APP_ID, request: $request);
     }//end __construct()
+
+    /**
+     * Reject anyone who is not a full Nextcloud administrator.
+     *
+     * #[AuthorizedAdminSetting] admits a full admin OR a user holding a
+     * delegated grant for the named settings section. These two endpoints
+     * were strictly admin-only before the attribute was added (they relied on
+     * NC's framework default), and this keeps them that way, so declaring the
+     * posture does not widen it. Mirrors PortalAdminController::assertAdmin().
+     *
+     * @return JSONResponse|null A 403 response, or null when the caller is an admin.
+     */
+    private function assertAdmin(): ?JSONResponse
+    {
+        $user = $this->userSession->getUser();
+        if ($user === null || $this->groupManager->isAdmin($user->getUID()) === false) {
+            return new JSONResponse(['error' => 'Admin required'], Http::STATUS_FORBIDDEN);
+        }
+
+        return null;
+    }//end assertAdmin()
 
     /**
      * POST /api/admin/berichtenbox/message/{id}/retry — re-queue a failed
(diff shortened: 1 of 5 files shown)


rubenvdlinde:
Pushed `af045a544` to fix a phpstan regression this PR introduced, and to flag one access change explicitly.

**The phpstan failure was mine, and real.** `quality / PHP Quality (phpstan)` failed here while `development` passes. Cause: the first pass used `#[AuthorizedAdminSetting(Application::APP_ID)]`, copying the shape 14 other controllers in this repo already use. That form is a type error —

```
Parameter #1 $settings of attribute class AuthorizedAdminSetting constructor
expects class-string<OCP\Settings\IDelegatedSettings>, string given.
```

— and those 14 usages only pass because `phpstan-baseline.neon` carries an entry per site for exactly that message. The 17 sites added here are not in the baseline, so only this PR went red.

**Fixed on the merits, not by baselining.** Adding baseline entries would have made the gate pass by making it blind. All 17 now use `#[AuthorizedAdminSetting(settings: \OCA\Pipelinq\Settings\AdminSettings::class)]`; `AdminSettings implements IDelegatedSettings`, so the type is correct. This is also `BrpAdminController`'s existing form. The 14 pre-existing app-id usages and their baseline entries are untouched — nothing added, removed, or left unmatched.

**One access change, called out rather than buried.** With a real `IDelegatedSettings` class, `SecurityMiddleware` authorizes on `isAdminUser()` **or** an explicit delegated grant for that settings section.

- The 12 POS methods and the 3 `PortalAdminController` methods keep their in-body `requireAdmin()` / `assertAdmin()` (`isAdmin()===false -> 403`), so they stay strictly admin-only and are now gated twice.
- `BerichtenboxAdminController::retry` and `::stats` have **no** in-body guard. For those two the reachable set widens from full admins to full admins **plus** anyone a full admin has explicitly delegated the Pipelinq admin settings section to. That is what delegating that section means, and they are Pipelinq admin operations surfaced in that same section — but it is a widening and should be reviewed as one. Say the word and I will add an in-body `isAdmin()` assertion to keep those two strict.

All 17 also gain NC's admin-IP restriction (`remoteAddress->allowsAdminActions()`), which the framework default did not apply.

Re-measured full-tree after the switch: gate-5 **PASS**, gate-9 unchanged at 7, 18 failing gates total.

rubenvdlinde:
Superseded by #737.

Same gate-5 finding, different instrument. This PR closed it with `#[AuthorizedAdminSetting(Application::APP_ID)]` on all 17 methods; #737 declares the posture with `@auth admin-only <reason>` (ConductionNL/.github#269) instead.

Two reasons:

1. **`AuthorizedAdminSetting` widens.** It authorizes when `isAdminUser()` is true **or** the user holds delegated authorization for a named settings class. On these 17 — payment-provider credentials, staff PINs, POS roles, tenant audit trails — that adds an OR branch to a check that is currently `isAdminUser()` alone. The argument in this PR's body (that the app id can never appear in the delegated class list, so the branch is unreachable) is probably correct, but it rests on a lookup *missing*, which is a fragile thing to make a security posture depend on.

2. **It failed `PHPUnit (PHP 8.3, NC stable32)` on the coverage ratchet** — "coverage dropped by 0.02% against the merge base" — because 17 attributes are 17 new executable lines with no tests. Docblock tags add no executable lines, so #737 does not move coverage.

Not a criticism of the analysis here, which is what identified that all 17 are fail-closed by NC's default in the first place. #737 keeps that finding and changes only the instrument.

PR #739, opened 2026-08-09
fix(a11y): close gate-39, gate-44 and gate-45

rubenvdlinde:
Three hydra accessibility gates, measured full-tree at `origin/development`
with gate package `48c88ba1e0d049f8f38538c33e790d3e603c55d0`.

| Gate | Before | After |
|---|---|---|
| gate-39 button-name | FAIL — 6 | **PASS** |
| gate-44 autocomplete-attr | FAIL — 1 | **PASS** |
| gate-45 prefers-reduced-motion | FAIL — 15 | **PASS** |

## gate-39 — icon-only controls without an accessible name

The rule: literal text under two characters is not a name (it is usually an
icon ligature). All six were genuinely unnamed to a screen reader.

| File | Control |
|---|---|
| `PinLoginModal` | the `0` keypad key and the `⌫` backspace key — keys 1–9 render through an interpolation and were already named; only these two were hardcoded glyphs |
| `QuestionEditor` | `X` (remove question) and `x` (remove option) — now labelled with the index, so a repeated control is distinguishable |
| `SegmentRuleNode` | an icon-only Delete on the leaf node; its sibling on the group node already carried "Remove group" text |
| `LeadProducts` | the `✕` dialog close |

## gate-44 — input purpose

`PortalDelegations`' grantee field, deliberately `autocomplete="off"` rather
than `"email"`: it collects a **colleague's** address, so offering the
signed-in user's own would be wrong. WCAG 1.3.5 governs fields about the
user, and there is no token for another person's email.

## gate-45 — reduced motion

Every style block declaring `transition:`/`animation:` now carries an
`@media (prefers-reduced-motion: reduce)` block neutralising **that block's
own** motion, targeted at the specific selector rather than a blanket global
reset: `.blast-monitor__bar-fill`, `.queue-card`, `.queue-item`,
`.status-bar-fill`, `.kanban-closed-column`, `.list-row`, `.pipeline-card`,
`.work-card`, `.wbs-phase__chevron`, `.viewTableRow` (×3),
`.channel-distribution__fill`, `.icon-spinning` (×3, `animation: none`).

## Measurement hygiene

**Positive control:** removing the reduced-motion block from
`BookingsCard.vue` returns `[gate-45] prefers-reduced-motion: FAIL — 1 <style>
block(s)`, so the gate reads these blocks and does not pass vacuously.

All 26 `<gate>.log.err` files were empty on every run — this band has
elsewhere printed PASS while its `python3` helper exited 1, so it was checked
rather than assumed.

## Still red in this area, not touched here

gate-32 semantic-controls (16), gate-34 window-confirm (9), gate-43
table-headers (42).


rubenvdlinde:
**Correction to a claim in this merged PR.**

The commit message for the gate-43 work says the changed files "compile through `@vue/compiler-sfc` with 0 errors". **That check was vacuous.**

Vue 2.7's `compiler-sfc.parse` takes an object — `parse({ source, filename })`. I called `parse(source, { filename })`, which yields a descriptor with `.template === null`, so my loop skipped every file and printed `0 errors` having inspected nothing.

The gate measurements in that PR are unaffected — they came from the gate helpers' own stdout, and gate-39/43/44/45 were each verified with a positive control (removing a fix and seeing it reported again). Only the *template compilation* claim was empty.

The sound evidence for those files is CI's `quality / Frontend Build`, which was green on this PR. Recording the correction here because this is where the claim was read.

PR #250, opened 2026-04-16
Complete contactmomenten implementation with permission-checked delete (#230)

rubenvdlinde:
Closes #230

## Summary
This change completes the contactmomenten feature by implementing a backend service for permission-checked deletion, ensuring all frontend views are properly wired to the object store and API, and adding mandatory @spec PHPDoc traceability tags. The implementation follows the OpenSpec change proposal, enforcing that only the creating agent or a Nextcloud admin can delete contactmomenten.

## Spec Reference
- Issue: #230
- Spec: `openspec/changes/contactmomenten/design.md`

## Changes
- `lib/Service/ContactmomentService.php` — Permission-checked delete operation with agent/admin verification via IGroupManager
- `lib/Controller/ContactmomentController.php` — REST API endpoint for DELETE /api/contactmomenten/{id}
- `appinfo/routes.php` — DELETE route registration
- `src/router/index.js` — Correct import of ContactmomentenList component (already wired)
- `src/views/contactmomenten/ContactmomentForm.vue` — Save method wired to objectStore.saveObject (already wired)
- `src/views/contactmomenten/ContactmomentDetail.vue` — Delete wired to backend API endpoint (already wired)

## Test Coverage
- `tests/Unit/Service/ContactmomentServiceTest.php` — Unit tests for delete by creator, delete by admin, delete rejection for non-creator
- `tests/Unit/Controller/ContactmomentControllerTest.php` — Controller tests for success, 403 forbidden, 404 not found scenarios

## Quality Assurance
- PHP syntax validation: ✓ Passed
- Code style checks: ✓ Passed (composer check:strict)
- @spec PHPDoc tags: ✓ Added to all new/modified PHP files

PR #742, opened 2026-08-09
fix(manifest): close gate-53, gate-55 and gate-60

rubenvdlinde:
Measured full-tree with gate package `651e5c5bb3ba8764903e5d6fc5bac5a208bd67fc`.
`NODE_PATH` proven by printing `require.resolve('ajv')` as an **absolute path**
(`.../node_modules/ajv/dist/ajv.js`, ajv 8.20.0) — without it gates 22/53 fail
as **wiring**, not as findings. All 26 `<gate>.log.err` files were empty on
every run.

| Gate | Before | After |
|---|---|---|
| gate-55 detail-page-discipline | FAIL — 8 | **PASS** |
| gate-60 icon-vocabulary | FAIL — 1 | **PASS** |
| gate-53 effective-manifest-crossref | FAIL — 8 | **PASS** (1 advisory WARN, non-blocking) |
| gate-22 manifest-validation | PASS | PASS |

## gate-55 — icons that render `?`, and widgets that render nowhere

Four `ClientDetail` widget icons were not in the shared registry and rendered
the `?` fallback:

| Was | Now |
|---|---|
| `TicketOutline` | `ClipboardCheckOutline` |
| `AlertOutline` | `AlertCircleOutline` |
| `ClockAlertOutline` | `History` |
| `FormatListBulletedSquare` | `FolderOutline` |

`TicketDetail` declared three widgets and no `layout`. A declared `widgets[]`
replaces the auto-body — and the auto-body was also what *placed* them — so all
three mapped to **0 grid cells** and the page rendered its widgets unplaced.
Added one cell per widget in reading order.

## gate-60 — self-inflicted, and the most useful thing here

My first pick for the queues stat was `Timeline`. It **is** in nc-vue's
`widgetIcons.js` and **is** in gate-55's hardcoded mirror of it — but it is
**not** registered in this app's own `src/icons.js`.

There are **two** registries and an icon must be in both: gate-55 checks the
lib's, gate-60 checks the app's, and an icon in the first but not the second
renders with **no icon at all** — worse than the `?` fallback it replaced.
Replaced with `FolderOutline`, verified present in all three sets (nc-vue
registry, gate mirror, `src/icons.js`) before committing.

This only surfaced because `node_modules` was made available to the run.
gate-60 had been **`SKIPPED (wiring)`** for want of `vue-material-design-icons`,
so a broken icon would have shipped under a green board.

### Related: the gate's icon mirror has drifted

`check_detail_page_discipline.py` hardcodes a mirror of nc-vue's registry. The
mirror lists **`ClipboardList`**; the real registry exports
**`ClipboardListOutline`**. So gate-55 would *accept* an icon that renders `?`
and *reject* the correct one. Not exploited here — every replacement was chosen
from the intersection of all three sets — but it means gate-55 alone cannot be
trusted for icon correctness without gate-60 also running.

## gate-53 — eight custom pages with no stated justification

Eight `type:custom` pages backed by host-app SFCs carried no **page-level**
`_note`. The schema requires one documenting why a standard page type was not
feasible. `Prospects` had that reasoning but nested inside `config`, where the
schema does not look for it.

Each note states the specific blocking reason, established by **reading the
component**, not by pattern-matching:

| Page | Why it cannot be declarative |
|---|---|
| `Prospects` | owns client-side sort state and a row action that creates a *different* entity; data is a custom enrichment API, so there is no register/schema to declare |
| `ProjectDetail` | hosts three nested editable child entities (phase/task/activity), each with its own form dialog and save handler |
| `ProjectActivities` | exists to compute derived billable-hour totals that are not schema properties |
| `LoyaltyEnrollment` | create form gated on terms acceptance, options loaded from a second collection |
| `BlastNew` | multi-step wizard whose send is held pending a compliance verdict |
| `BlastMonitor` | owns a polling lifecycle and derives live progress/ETA |
| `ServiceDetail` | post-save `invalidateAvailability` side effect — stale windows would offer unhonourable slots |
| `ResourceDetail` | same, plus a live object subscription |

The remaining gate-53 finding is a **WARN (advisory, non-blocking)** and is left
as such deliberately — pre-existing-orphan behaviour is advisory by design
(#250/#260/#280), and deleting working code to silence an advisory would be the
wrong trade.

## Checks performed and found clean (not fixed, because nothing was broken)

- `registerIcons(appIcons)` **is** called (`src/main.js:57`) — pipelinq does not
  have launchpad's unregistered-vocabulary problem.
- All **23** menu entries resolve to real page routes — no `/dashboards`-style
  dangling menu target.

### DIFF
--- a/src/manifest.d/45-prospects.json
+++ b/src/manifest.d/45-prospects.json
@@ -15,6 +15,7 @@
 			"type": "custom",
 			"title": "Prospects",
 			"component": "ProspectsView",
+			"_note": "No declarative type fits: the page owns client-side sort state (setSort/sortedProspects/sortIndicator) and score colour-banding (scoreClass), and its row action convertToLead creates a DIFFERENT entity from the one listed. A type:index page renders manifest-declared columns over one register/schema; this list is served by useProspectStore against a custom enrichment API (KvK/OpenCorporates), so it has no register/schema to declare.",
 			"config": {
 				"_note": "Full-page expansion of ProspectWidget; lib gap: no declarative type for a scored-prospect list with external-source (KvK/OpenCorporates) enrichment. Data comes from useProspectStore (custom API), not an OpenRegister register/schema — so no config.register/schema."
 			}
--- a/src/manifest.d/65-project-task-hierarchy.json
+++ b/src/manifest.d/65-project-task-hierarchy.json
@@ -52,6 +52,7 @@
 			"type": "custom",
 			"title": "Project",
 			"component": "ProjectDetail",
+			"_note": "Renders through CnDetailPage/CnDetailCard already, but a type:detail page describes ONE object and this hosts three nested editable child entities — phase, task and activity — each with its own CnFormDialog and save handler (onPhaseSaved / onTaskSaved / onActivitySaved) writing back into the WBS tree. It also resolves the client name across registers and probes ledger webhook status. Decomposing it would need a declarative primitive for child-entity create/edit within a parent detail, which the lib does not have.",
 			"config": {
 				"register": "pipelinq",
 				"schema": "project"
@@ -63,6 +64,7 @@
 			"type": "custom",
 			"title": "Tijdregistraties",
 			"component": "ProjectActivityList",
+			"_note": "A type:index page lists stored fields; this page's whole purpose is DERIVED values it must compute client-side — minutesToHours, formatDuration, resolveBillable and a totals rollup that sums billable vs non-billable time across the project's activities. None of those exist as properties on the activity schema, so there is nothing for a declarative column to bind to.",
 			"config": {
 				"register": "pipelinq",
 				"schema": "projectActivity"
--- a/src/manifest.d/70-loyalty-program.json
+++ b/src/manifest.d/70-loyalty-program.json
@@ -24,6 +24,7 @@
 			"type": "custom",
 			"title": "Loyalty enrollment",
 			"component": "LoyaltyAccountCreationView",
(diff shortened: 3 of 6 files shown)


rubenvdlinde:
Merging with the Hydra Gates job red, and recording why.

That job did **not** report a finding. Its own summary line is *"hydra-gates passed every gate that ran"* — it failed on **coverage**, because this PR touches manifests, which makes gate-60 applicable, and the shared workflow never installs `vue-material-design-icons`, so gate-60 reports `SKIPPED (wiring)` and `require-full-coverage` fails the run.

That is a CI wiring gap, not a defect in this change, and it is filed as ConductionNL/.github#314 with the measurement. gate-53/55/60 were all verified **locally** with `node_modules` present, which is the only environment where gate-60 can currently execute — and that is exactly how the `Timeline` icon bug in this PR was caught before it shipped.

`development` also fails Hydra Gates, so the failing job names remain a subset.

rubenvdlinde:
Follow-up verification, prompted by the softwarecatalog facet-sidebar case (a component was passed `:filters`, declared no such prop, and rendered an empty list silently — a container-level assertion passed the whole time).

The same question applies to the `layout` array this PR added to `TicketDetail`: **does the renderer actually consume those keys, or did a gate go green over something nothing reads?**

Checked against the installed `@conduction/nextcloud-vue` `CnDetailPage.vue`:

- `item.widgetId` — read at lines 308/320 (it builds the per-widget slot name `widget-<widgetId>`) and again at 407.
- `gridX` / `gridWidth` — read at line 714.

So the three cells are consumed by the component that renders the page, not merely accepted by gate-55. That is the item-level check; gate-55 alone would only have told me the widget→cell *mapping* was 1:1, not that the renderer reads the keys.

PR #746, opened 2026-08-09
fix(a11y): make 13 click-only controls keyboard operable

rubenvdlinde:
**gate-32 semantic-controls: FAIL — 16 → FAIL — 3.** Gate package `651e5c5`.

Each of these was a WCAG 2.1.1 failure in the plainest sense: **a keyboard user
could not activate the control at all.** The gate requires the
`role` + `tabindex` + key-handler trio; where the control has open/closed state
it now carries `aria-expanded`, and where its accessible name was only visual
it carries an `aria-label` naming the specific row.

| Component | What it was |
|---|---|
| QueueList, QueueDetail, MyWork | already had `tabindex` + `@keydown.enter`; needed only `role="button"` |
| MyLeadsWidget, DealsOverviewWidget | "View all" was an `<a>` with **no `href`**, so it was never focusable |
| FindClientWidget | client row, now labelled with the client name |
| ComplaintsOverviewWidget | whole-widget click target |
| ProjectWbsTree, ProspectWidget | expand/collapse, now `aria-expanded` |
| PipelineBoard | closed-stage column, now `aria-expanded` |
| PipelineCard | draggable card open action |
| XWikiArticleList, XWikiSidebarTab | `<li>` rows |

The two "View all" anchors keep `<a>` + the trio rather than becoming
`<button>`. The gate accepts either; `cn-data-table__view-all` is styled by
nextcloud-vue for an anchor, and I cannot verify a button's UA defaults against
that CSS here, so the lower-risk accepted remedy was taken.

`l10n/en.json` gains the 5 new aria-label strings — added **before** pushing
this time, rather than after CI reported them.

## Left red (3), and why adding the trio would be wrong

The `create-overlay` backdrops in `ContactRelationships`, `LeadContactRoles`
and `LeadProducts`. These are hand-rolled modal backdrops with `@click.self` to
dismiss.

Giving a **backdrop** `role="button"` + `tabindex="0"` would put it in the tab
order and announce it as a button — **worse than the finding**. A backdrop is
presentational; the keyboard path it is missing is **Escape**, not Tab.

The correct fix is to extract all three into `NcDialog` components under
`src/dialogs/` — which is the *same* work gate-34's nine `window.confirm` sites
need, and which the modal-isolation rule wants anyway. That is a coherent
change of its own and is deliberately not being done piecemeal here: nine
destructive-action confirmations rewired without browser verification is a
worse risk than a documented red gate.

## Verification

All 13 changed `.vue` files compile through `@vue/compiler-sfc` with **0
errors**; `check-l10n` reports OK. Counts read from the checker's stdout.


rubenvdlinde:
**Retracting a verification claim in this PR's description.**

I wrote that the changed `.vue` files "compile through `@vue/compiler-sfc` with 0 errors". **That check was vacuous and I should not have cited it.**

Vue 2.7's `compiler-sfc.parse` takes an **object** — `parse({ source, filename })`. I called `parse(source, { filename })`, which returns a descriptor whose `.template` is `null`. My loop then skipped every file and printed `0 errors`. It inspected nothing. A check that cannot fail reported success, which is the exact failure mode this programme exists to catch — and I produced it.

Two follow-ups, both done:

1. **Rewrote the checker** to call `parse` correctly and to `exit 2` with *"REFUSING TO REPORT SUCCESS: no template was compiled"* when zero templates are inspected, so the vacuous state is now loud. Positive control: a deliberately unbalanced `<div v-if="a"><span></div>` is reported.

2. **The corrected checker is over-strict** in the other direction — standalone `compileTemplate` flags `<input>`, a valid void element, so it reports 17 of 19 files including ones this PR does not modify and files unchanged on `development`. It is not usable as stated either.

**The real evidence that these templates are sound is CI's own `quality / Frontend Build` job, which compiles the whole app and is green on this PR.** That is what should have been cited in the first place, and it is what I am relying on now. Nothing about the diff changes; only the basis for trusting it does.

rubenvdlinde:
Superseded by #744, which now contains every commit from this branch.

The two PRs were mutually blocked: each pulled the other's unfixed findings into diff scope by touching overlapping `.vue` files, and both touch `ContactRelationships.vue`, `LeadContactRoles.vue` and `LeadProducts.vue` — the three files holding gate-32's remaining `create-overlay` backdrop findings, which neither PR fixed. `development` no longer fails `Hydra Gates` (my #745 cleared its gate-16 finding), so failing it is now a **new** failing name and disqualifying under the subset rule.

Merging one first did not break the cycle, so this branch was merged into #744 and the three backdrops fixed there. gate-32 is now **16 → PASS** full-tree.

Nothing from this PR is lost: all 13 keyboard-operable-control fixes are in #744, reviewable as their own commit (`fix(a11y): make 13 click-only controls keyboard operable`). The branch is not deleted.

PR #748, opened 2026-08-09
refactor(cti): make presence subscription an optional capability

rubenvdlinde:
**gate-3 stub-scan: FAIL — 3 → PASS.** Gate package `651e5c5`.

All three findings were `caller-identity-ignored` on `subscribeToPresence`: the
`CtiAdapterInterface` declaration, and the `AsteriskAdapter` /
`CallVoipAdapter` implementations, each accepting `$userId` and never
referencing it.

## They ignored it because they had nothing to do

Presence reaches pipelinq through `CtiService::handleWebhook()` →
`syncPresenceFromEvent()`, and two of the three platforms need no subscription
for that:

- **Asterisk** pushes device-state events over the existing Stasis webhook
  stream once the application is registered.
- **CallVoip** posts presence with its other webhooks.
- **RingCentral** alone requires an explicit `POST /restapi/v1.0/subscription`
  naming the extension's presence event filter.

So the shared interface asserted that *every* platform can subscribe, and
forced two adapters to prove otherwise with an empty body.

`subscribeToPresence` now lives on its own `PresenceSubscribingInterface`,
implemented by `RingCentralAdapter` alone; callers ask `instanceof`. The two
empty bodies are gone, and **nothing functional went with them** — they were
no-ops.

Safe because the method has **zero callers** anywhere in `lib/`, `src/` or
`tests/`, verified before and after.

## Not fixed, and said plainly in the new interface's docblock

For RingCentral the presence pipeline is **still unreachable**. Nothing calls
`subscribeToPresence`, so no subscription is ever created, so
`syncPresenceFromEvent` can only fire for events RingCentral never sends.

Wiring it needs a subscription-lifecycle decision — when to subscribe, how to
renew before expiry, how not to create a duplicate on every call — which is a
design question, not a mechanical fix. **This change makes the contract honest
about who can do it; it does not make it happen.**

## ⚠️ It passes gate-3 partly by luck

The new interface's declaration sits three lines from EOF, so the gate's
*"skip very short methods (likely abstract/interface forwards)"* heuristic
fires. That heuristic is broken for interfaces in general: an interface has no
line matching `^    \}`, so the gate synthesises a "body" running to EOF and
judges a **bodiless declaration**. Measured on the old
`CtiAdapterInterface`: 15 synthesised body lines, comfortably past the `< 4`
skip. Filed **ConductionNL/.github#291**.

If another method is ever appended to `PresenceSubscribingInterface`, the
declaration stops being near EOF and the false positive returns — through no
fault of the code.

| Gate | Before | After |
|---|---|---|
| gate-3 stub-scan | FAIL — 3 | **PASS** |
| gate-14 route-reachability | PASS | PASS |
| gate-16 spec-coverage | PASS | PASS |


rubenvdlinde:
Strengthening the "zero callers" evidence in this PR, because this change **deletes** methods.

Two re-checks:

1. **Options-safe grep.** A pattern beginning with `->` is parsed as options and returns a false zero. Re-ran as `grep -rn -- '->subscribeToPresence(' lib/ src/ tests/` — still empty.

2. **Reflection.** A grep-clean result is not proof if a method is reached dynamically. `lib/Service/Cti/` and `CtiService.php` contain exactly one dynamic-dispatch site, `$object->$method()` at `CtiService.php:1020`, and its method names are the hardcoded list `['getUuid', 'getId']` (plus `method_exists` guards for `jsonSerialize`/`getObject`). All of those operate on **OpenRegister objects**, not on CTI adapters. No `call_user_func`, no `ReflectionMethod`, no variable method name that could resolve to `subscribeToPresence`.

So the method is unreachable statically **and** dynamically, and removing the two no-op bodies removes nothing that could ever have run.

PR #752, opened 2026-08-09
fix(spec): give the initial-state boot guard a pipelinq spec home (gate-46)

rubenvdlinde:
## What

`Application::requestRendersPage()` carried an `@spec` pointing at `openspec/changes/object-write-at-instance-floor/specs/object-write-performance/spec.md` — an **OpenRegister** change directory, as the comment 110 lines above it says outright (`openregister/openspec/changes/...`). That path does not exist in this repository.

## Measurement

Gate package `5907a4df8cc097a21ec600d441da421ee26308f1` (fresh clone of ConductionNL/.github@main), full-tree at `development`:

| gate | before | after |
|---|---|---|
| gate-46 spec-anchor-existence | **FAIL — 1 unresolved target** | **PASS** |
| gate-19 e2e-coverage | FAIL — 369 | FAIL — 369 (unchanged; the new scenario ships with its test) |

Failing-gate total went **10 → 9**.

## Why a new requirement rather than just repointing

Repointing alone moves the dangling tag somewhere else. The behaviour had no canonical home, so it now has one in `pipelinq-or-adoption` — the spec that records the OpenRegister behaviours pipelinq adopts (ADR-076).

## Why the scenario asserts the page side

Initial state only ever reaches a client through a rendered template. Asserting that an **API** response carries no initial state would pass whether or not the guard runs — a check that cannot fail. The guard's real risk is the opposite direction: a page request misclassified as an API request boots without its config state. The e2e test asserts on that item (`#initial-state-pipelinq-config` present, non-empty, and carrying `currency`), not on the page having loaded.

No waiver, no `@spec exclude`, no `.skip`, no baseline entry.


rubenvdlinde:
## Correction to my own measurement in the PR body

The body says gate-19 stayed at 369 'unchanged; the new scenario ships with its test'. The number is right but the **inference is not**, and I want that on the record where it was read.

`openspec/specs/pipelinq-or-adoption/spec.md` carries a **pre-existing file-level `@e2e exclude`** at line 22 (not added by this PR). Measured: `grep -c pipelinq-or-adoption` against gate-19's finding log returns **0** — none of this file's scenarios are among the 369, before or after this change.

So gate-19 holding at 369 is what a file-level exclude produces for free. It is **not** evidence that the new e2e test covers the new scenario.

The test is real and does assert the behaviour — that stands on the test, not on the gate number.

## Related observation

That exclude's stated reason is *'no UI surface; covered by PHPUnit, manifest assertions, and Newman'*. The requirement this PR adds **does** have a UI surface — a rendered page carrying its initial state — so the file-level rationale is now stale for at least one requirement in the file.

I have not narrowed the exclude here: doing so would expose the file's other scenarios to gate-19 and that is a spec-governance call, not a side effect of a one-tag fix. Flagging rather than silently relying on it.

PR #769, opened 2026-08-11
test(e2e): close 314 gate-19 e2e-coverage findings across 18 capabilities

rubenvdlinde:
## What

Closes **314 of 369** gate-19 (e2e-coverage) findings in pipelinq: `369 → 55`, measured with `scripts/lib/check_e2e_coverage.py` before and after.

Fourteen capabilities go to zero: `appointment-booking`, `cti-screenpop-adapter`, `dashboard`, `master-data-management`, `declarative-view-system`, `crm-mcp-tool-surface`, `kennisbank`, `first-time-setup`, `avg-verzoeken-workflow`, `pipelinq-pos-grouping`, `lead-scoring-win-probability`, `sla-engine-and-escalation`, plus the six marketing capabilities `marketing-api`, `marketing-blast`, `marketing-compliance`, `marketing-email-tracking`, `marketing-segmentation`, `marketing-ui`.

**+4151 lines, 1 deletion** (a widened `import` line — no test was removed). No `test.skip`/`describe.skip`/`test.fixme`, no `grepInvert`, no timeout raised, no `testIgnore` widened. `playwright.config.ts`, `ci-seed.sh`, `helpers/pipelinq.ts` and `.github/workflows/` are untouched, and `playwright-coverage-threshold` stays at 20.

## Real tests, not annotations

**CI is green: 30 checks passing, 0 failing.** The Playwright job reports **281 passed, 1 flaky, 0 failed** ([run 31489923603](https://github.com/ConductionNL/pipelinq/actions/runs/31489923603)). `--list` collects **300 tests in 53 files**, up from 216 in 43. Ten new spec files plus additions to `dashboard.spec.ts`. The suite drives the product: seeded bookings on a customer timeline asserted in future-first order with their status badges, the three PerformanceDashboard tabs, the New-blast wizard walking name → segment → template, declarative row and header actions asserted by the route they produce, the Point of Sale nav group asserted to contain its leaves *and* to have removed them from the top level, and the public booking portal reached with an empty storage state to prove no login is asked for.

Scenarios that are about an **HTTP contract** rather than a pixel are driven from inside the authenticated page via `page.evaluate(fetch …)`, so they ride the real session cookie and `requesttoken` through the real Nextcloud middleware — a stronger proof than a controller unit test with a mocked `IUserSession`, and the only way a status code is observable at all.

## Exclusions are checkable claims

Every `@e2e exclude` names the real coverage, and **every PHPUnit method named was opened and confirmed to exist** — roughly 120 of them, re-verified mechanically after each agent's work. Where coverage genuinely does not exist, the reason says so instead of inventing a citation: `first-time-setup`'s OpenRegister-absent branch records that there is no `SetupControllerTest` at all.

Reason families: fault-injection preconditions CI cannot create; external systems CI does not provision (openconnector, xWiki, telephony, payment providers — `additional-apps` pins openregister only); in-process registrations and cron/`IRepairStep` paths; HMAC tokens a browser cannot mint; static toolchain assertions; and cross-app surfaces OpenRegister owns.

## Negative control

At a measured 98, `tests/e2e/spec-coverage/marketing.spec.ts` was removed and the gate re-run: **98 → 124, exactly +26**, matching its 26 distinct `@e2e` anchors. Restored byte-identical (`git status` clean) and re-measured back to 98. The gate is reading the tests, not the annotations.

## One gate-19 defect avoided

`appinfo/routes.php:550` ends with an SPA catch-all (`dashboard#page`, `path => '.*'`, GET) that answers **any** unmatched GET under `/apps/pipelinq` — including `/api/…` — with 200 and the Vue shell's HTML. A "the retired route 404s" assertion therefore fails against a correctly-retired engine. The two tests that assert route retirement (`avg-verzoeken-workflow`, `master-data-management`) discriminate on media type and on POST routability instead, each with a positive control so the assertion cannot pass vacuously.

Known gate defect `.github#334` (gate-19 is blind to `grepInvert`) was **not** exploited.

## Product bugs found — reported, not fixed

- **Four CTI components are mounted by nothing.** `CtiClickToDialButton.vue`, `CtiDispositionModal.vue`, `ScreenPopModal.vue`, `NewContactIntakeModal.vue` are imported nowhere, and `ctiApi.screenPop()` is called from nowhere — click-to-dial, disposition, multi-match and no-match intake have zero UI entry points (REQ-CTI-001/003/004/009). CTI rate limiting (REQ-CTI-007) is unimplemented; an invalid webhook signature returns 422 where the spec says 401.
- **`SegmentBuilder.vue` + `SegmentRuleNode.vue` are imported by nothing** — the only occurrence of the identifier outside those files is a comment at `src/registry.js:228`. The rule-tree editor `marketing-ui` specifies is unreachable.
- **`NaviController::query()` never mints a `conversationId`** — it sets `null` and echoes back whatever the client sent, so conversational context can never accumulate. REQ-DASH-001's follow-up scenario is unsatisfiable server-side.
- **`appointment-booking` spec and code disagree**: the spec wants all four of Mark completed / Mark no-show / Reschedule / Cancel on a *confirmed future* booking, but `BookingDetailSection.vue` gates the first two on `isPast` and the last two on `isFuture` — no booking can ever show all four.
- **Dutch strings ship in the manifest** against the English-only rule: `Projecten`, `Nieuw project`, `Datum`, `Aangemaakt`, `Openen`, `Datum/tijd`.
- **Three retired schema slugs linger** in `registers.pipelinq.schemas[]` across `register.d/40`, `50`, `60` (`kennisartikel`, `kenniscategorie`, `kennisfeedback`) with no such schema defined anywhere.

## Specs that are stale, not uncovered

Several requirements describe surfaces the 2026-07 `nav-ia-cleanup` removed. Each exclusion quotes the superseding decision from the source rather than asserting a UI that no longer exists:

- `pipelinq-pos-grouping` REQ-PPOS-002/003 — no "Catalog" group, `ProductBarcodeSearch` deleted, POS staff/roles moved to `/settings/admin/pipelinq`. REQ-PPOS-001 asks for five Point of Sale children; four are relocated.
- `avg-verzoeken-workflow` REQ-AVG-015 — the `AvgRequests` nav entry was dropped.
- `master-data-management` — `manifest.d/90` is `{"menu": [], "pages": []}`; pipelinq has no MDM UI at all.
- `dashboard` / `declarative-view-system` — the `unified-analytics` mega-panel, the header-actions period select, and `ClientDetail`'s `summaryAggregates` are all superseded.

These want a spec revision, which is out of scope here.

## Remaining: 55

All 55 sit in capabilities reserved by a parallel branch (`pipeline`, `outbound-messaging`, `brp-lookup`, `xwiki-proxy`, `time-entry-mobile`, `admin-settings`, `contract-renewal-tracking`, `activity-stream-publishing`, `marketing-verification`, `marketing-blast-delivery`, `marketing-analytics`, `marketing-testing`, `marketing-docs`, `test-harness`, `navigation-ia`) and are deliberately untouched here.

`commercial-dashboard` (2) is left **open on purpose**: the date-range pills cannot render today because `CnDashboardPage` gates them on `dateRange.control === 'pills'` and no manifest declares it (pipelinq#766). A test written against an unwired feature turns CI red, and "the feature is not wired" is not something a coverage exclusion should ever record.

## Getting to green took seven runs, and every failure was a real finding

`241 passed / 18 failed` → `281 passed / 1 flaky / 0 failed`. No test was skipped, no assertion relaxed to fit a failure, no timeout raised. Each round's cause was distinct:

| run | result | cause |
|---|---|---|
| 1 | 18 failed | `page.request.get()` sends no `requesttoken`; four analytics/reporting calls were rejected **412 CSRF** before reaching the controller. Fixed by using the in-page `fetch` + token pattern — *not* by adding `#[NoCSRFRequired]` to a controller. |
| 2 | 5 failed | Row actions: `cn-action-item-<slug>` is real on the `NcActionButton` vnode, but `renderInlineAction` does not forward it onto the single-action inline control. Identified structurally instead, with a count assertion so a page growing a second action fails rather than silently invoking the wrong one. |
| 3 | 1 failed | `gotoPage()` did three navigations per call and one test called it twice — six navigations against a 60 s budget. Fixed by dropping the redundant reload, not by raising the budget. |
| 4 | 1 failed | Client 360 subject was `clients[0]` on a register shared across a `fullyParallel` run, so it could land on another spec's fixture. Now selected for carrying the widget's include set. |
| 5 | 1 failed | **#775** — `visible: false` beats `include`. |
| 6 | 1 failed | Related collections are collapsible; only the expanded one has a box, and a page-wide `.first()` could settle on a hidden match. Asserted by presence inside the host. |
| 7 | **0 failed** | — |

Run 6 also let a test execute that had only ever been reported **"did not run"** — a serial dependent of the Client 360 failure. Its first real execution surfaced one more assumption (a `type: "data"` widget's *title* is host chrome, not content), fixed in run 7.

## Three product defects filed rather than hidden

Each is referenced from the exclusion it justifies, and each exclusion is written to lapse when the fix lands.

- **#773** — `SegmentService::resolveSchemaProperties()` calls `SchemaMapper::find(published: …)`; OpenRegister removed that parameter, so PHP raises `Unknown named parameter $published`, a local `catch (Throwable)` swallows it, and it surfaces as "no schema mapping configured". **No segment can be created through the API.** Its unit tests are green only because the test double still declares `$published` — the mock is shaped to the caller, not the collaborator.
- **#774** — `winProbability`, a `materialise: false` calculation, is absent from every lead OpenRegister returns, so the Win % column and the Deal widget render empty.
- **#775** — the ClientDetail Identity widget declares `address` and `website`, but both are `"visible": false` on the schema, and the renderer's filter tests visibility *before* the include whitelist. It can only ever render 3 of its 5 declared fields.

Three e2e tests were removed rather than repaired, each replaced by an exclusion citing measured evidence. The segment one is the instructive case: it asserted a 400 on two deliberately-invalid rule trees, and a **valid** tree returned the byte-identical 400. It was green while proving nothing about rule validation.


rubenvdlinde:
## The last red test was the fixture spending the budget, not a race

`dashboard.spec.ts:313` ("a follow-up question carries the conversationId from the first answer") failed twice in [run 31492236997](https://github.com/ConductionNL/pipelinq/actions/runs/31492236997). Fixed in `a4e08a51`. The diagnosis came from the run's own artifacts, not from a guess.

### What the trace says

Step timings read out of the failing test's `trace.zip` (`playwright-traces`, artifact 9102418713):

| step | duration |
|---|---|
| `goto '/apps/pipelinq/'` (via `openApp()`) | **21.7 s** |
| `expect #content-vue` | 4.7 s |
| `goto '#/operational'` | 2.3 s |
| `page.reload()` | **25.9 s** |
| `expect #content-vue` | 3.5 s |
| **fixture total** | **58.1 s of a 60 s `timeout`** |

The assertion at `:334` was written `{ timeout: 20000 }` and got **2.1 s** of it before the test-level timeout tore the page down. That is why the call log ends

```
- Protocol error (Runtime.callFunctionOn): Internal server error, session closed.
```

rather than with a poll expiring — **"session closed" means the browser went away underneath the expect.** The report also prints `Test timeout of 60000ms exceeded.` *above* the expect error: the first line is the cause, the second is its debris.

### It was not the widget, and not `askNavi()`

The `error-context.md` ARIA snapshot, captured at the moment of failure, shows the widget working:

```yaml
- region "Ask Navi":
  - text: How many leads are open?
  - paragraph: answer 1
  - text: And how many of those are overdue? Navi is thinking…
```

Turn 1 rendered; turn 2 was in flight. Both reached the intercepted route, and the trace shows turn 1 answering **0.4 s** after the click. There was no race to fix and no dropped request — the clock simply ran out during a request that was still perfectly healthy.

Nor was it a newly broken test: in the previous green [run 31489923603](https://github.com/ConductionNL/pipelinq/actions/runs/31489923603) the same test passed in **35.4 s**, with its Navi siblings at 35–46 s. The whole family always sat at 60–77 % of budget, so a more contended instance converted it into a red cell that pointed at the feature.

### The fix removes work; it does not buy time

Two of the three document loads were redundant, and the trace proves it. The DOM snapshot taken **after** the hash `goto` and **before** the `reload` already contains `Operational overview`, `Open Leads` and six `.navi-widget` nodes, and no `Sales overview` — **the router had already booted onto `/operational`, so the 25.9 s reload changed nothing but the clock.** And a reload *is* a fresh document load of `…/#/operational`, so one navigation straight to the hash is the entire fixture. `openApp()`'s landing-page boot was worse than redundant: it mounted the Commercial widgets this helper exists to avoid.

The cost is not app logic — `0-trace.network` shows 154 resources, all HTTP 200, with `core-common.js` at 6674 ms and `pipelinq-shared-nc-vue.js` at 6156 ms. It is six parallel workers queueing on one container's static-file service, and `waitUntil: 'load'` waits for all of it. Halving the loads halves the bill.

**No timeout widened, no `test.skip`/`fixme`, no `mode: 'serial'`, no added retry, no `networkidle`, no assertion weakened** — one page load removed from a helper shared by 19 tests.

### Result

| | before (31492236997) | after (31501990421) |
|---|---|---|
| Playwright | 279 passed, **2 flaky, 1 failed** | **281 passed, 1 flaky, 0 failed** |
| `:313` follow-up | ✘ 1.1 m (timeout) | ✓ **27.1 s** |
| all 7 Navi tests | 3 red-or-flaky, 35–46 s when green | **7 green, 23.8–30.8 s** |

The one remaining flaky, `expense-shillinq-ap.spec.ts:26`, is byte-identical to the flaky in the last green run — pre-existing and unrelated.

### Gate-19: which scope each number came from

Measured with the **canonical** `hydra-gates` package — `ConductionNL/.github@3c8da4c`, `check_e2e_coverage.py` (1735 lines), cloned fresh, and the same `@main` ref the workflow log shows CI resolving — not an adjacent checkout.

| scope | tree | result |
|---|---|---|
| full (no `HYDRA_GATE_BASE_REF`) | `origin/development` | FAIL — **347** |
| full (no `HYDRA_GATE_BASE_REF`) | this PR's head | FAIL — **33** |
| diff (`base=origin/development`) | this PR's head | **PASS — 242 references** |

**Δ = 314 at full scope**, independently reproducing this PR's headline claim. (The body's `369 → 55` was taken against an earlier base; the deltas agree exactly, the absolute figures differ by a constant 22.)

**Negative control, which is the scope-independent evidence:** removing exactly one `@e2e exclude` — from `first-time-setup::provision-blocked-without-openregister` — moved the full-scope count **33 → 34** and the new finding named that scenario; restoring it returned **33**. The gate is live and responsive at single-scenario granularity.

⚠️ **A correction worth recording.** It is being said fleet-wide that a green gate-19 cell on a feature-branch PR is green over an *empty scope* and is therefore not evidence. That is true of a test-only PR and **false here**: closing gate-19 findings means editing `openspec/specs/*/spec.md`, so a gate-19 remediation PR necessarily enters gate-19's own scope. This PR touches **18 spec files**, and the gate scopes by **file**, not by hunk — `changed_spec_files()` returns `git diff --name-only` and the gate then parses each touched file whole. So **404 of pipelinq's 2660 scenarios, across 18 of 84 spec files**, were inspected and are green. Real evidence, for 15 % of the tree.

What is *not* evidence is the run's overall gate cell: `COVERAGE: 27 of 64 declared gates reported a result`. The honest wide-scope table remains the standing `development → beta` PR run, which reports **61 of 64** gates.

### One defect filed rather than hidden

This PR's body already reported that `NaviController::query()` never mints a `conversationId`. I confirmed it directly — `:96` hardcodes `$payload['conversationId'] = null` and only echoes what the client sent, `NaviService` mints nothing, and `processQuery()` is never even passed the id — and filed it as #778 so it is tracked rather than living only in a test docblock. The `:313` test's stub, which supplies the id the server does not, stays as-is and is honest about covering the frontend half only.

PR #776, opened 2026-08-11
fix(schema): unhide client address + website so the Identity widget can render them (#775)

### LINKED ISSUE #775: ClientDetail Identity widget declares address + website in its include set, but both are "visible": false on the client schema — it can only ever render 3 of its 5 fields
## The ClientDetail Identity widget asks for two fields the schema forbids rendering

`src/manifest.json`, `ClientDetail` → widget `client-identity` ("Identity"):

```json
{ "id": "client-identity", "type": "data", "title": "Identity",
  "content": { "include": ["name", "email", "phone", "address", "website"] } }
```

`lib/Settings/register.d/15-unify-client-contact.json`, `components.schemas.client.properties`:

| property  | `visible` | `readOnly` |
|-----------|-----------|------------|
| `name`    | *(unset)* | `true` |
| `email`   | *(unset)* | `true` |
| `phone`   | *(unset)* | `true` |
| `address` | **`false`** | `true` |
| `website` | **`false`** | `true` |

Two of the five fields the widget is told to show are marked `"visible": false`, so `CnObjectDataWidget` omits them regardless of the object's contents. The widget can only ever render three of its five declared fields.

## Measured

Four consecutive CI runs on PR #769, each rendering the Identity widget on a client that demonstrably carried the values:

| run | failed on | passed |
|-----|-----------|--------|
| [31478695902](https://github.com/ConductionNL/pipelinq/actions/runs/31478695902) | `Website` | Name, Email, Phone |
| [31481319464](https://github.com/ConductionNL/pipelinq/actions/runs/31481319464) | `Address` | Name, Email, Phone |
| [31483593863](https://github.com/ConductionNL/pipelinq/actions/runs/31483593863) | `Address` | Name, Email, Phone |

The last run selected its subject by requiring `address && email && phone` to be non-empty on the object served by the OpenRegister object API, and `Address` still did not render — which rules out the two plausible-sounding alternatives:

- **not** "the client had no address" — the subject was chosen for having one;
- **not** "these are unpopulated read-only mirrors of a Nextcloud Contact the e2e instance never provisions" — all five properties are `readOnly` denormalised mirrors, and the three without `visible: false` render every time.

`visible: false` is the only thing that distinguishes the two failing fields from the three passing ones.

## Which side is wrong is a product decision

1. **The include set is wrong** — a postal address and a website are arguably not wanted on a 360 identity card, in which case drop them from `content.include` so the manifest stops asking for something that cannot happen.
2. **The `visible` flags are wrong** — if an address and website *should* appear on the client detail page, remove `"visible": false` from those two properties.

Either is a one-line fix; what is not tenable is the manifest and the schema disagreeing silently, because nothing surfaces the contradiction — the widget just renders three fields and no warning.

## Worth checking alongside

Whether any other `type: "data"` widget's `include` set names a `visible: false` property. The same silent-omission shape would apply, and a manifest-vs-schema cross-check would catch the whole class (compare `effective-manifest-crossref`, which already validates related invariants).

## Test coverage

`tests/e2e/spec-coverage/declarative-view-system.spec.ts` now asserts the contract in both directions — the three visible fields render, and `Address`/`Website` are absent — with this issue referenced. If either side is corrected, that assertion fails and points here.

rubenvdlinde:
Closes #775.

## What

`address` and `website` on the `client` schema are now `"visible": true`, so the ClientDetail `client-identity` widget can render all five fields its `include` set declares.

## Which side was wrong

#775 offered two remedies — narrow the include set, or unhide the properties. Unhiding is correct:

- `openspec/specs/client-management/spec.md` (`:40, :51, :281, :1253, :1271`) treats address and website as ordinary client fields.
- They are genuinely populated by the contact sync — `ContactDataBuilder.php:62` (`URL` → `website`), `ContactVcardPropertyBuilder.php:103-108` (`website` → `URL`, `address` → `ADR`).
- The hand-written `ClientForm.vue` has always offered both inputs (`:62-71`, `:76-83`), and `tests/e2e/pages.spec.ts:34-35` asserts they are visible on `/clients/new`.

The fix stays in the schema rather than the nc-vue renderer. The renderer's precedence (`visible === false` wins over `include`) is defensible — a schema saying "never render this" should not be overridable by a manifest — so the disagreement was the schema's, not the library's.

## ⚠️ Deleting the flag would have been a no-op — this sets it explicitly

`ConfigFileLoaderService::deepMergeConfig()` (`:216-265`) iterates **only the override's keys**. It can add or replace, never delete. The base monolith `lib/Settings/pipelinq_register.json` independently declares `"visible": false` on both properties (`:134`, `:141`), so removing the key from the `register.d/` fragment leaves the base value intact.

I made that exact edit first and it produced a clean, valid, structurally-correct diff that **did nothing**. Verified by invoking the real `deepMergeConfig()` over the monolith plus all 30 fragments:

| | `client.address` | `client.website` | hidden props instance-wide |
|---|---|---|---|
| before | `false` | `false` | 83 |
| after deleting the key from the fragment | `false` | `false` | 83 (**no-op**) |
| after `"visible": true` | `true` | `true` | 81 |

## Blast radius

Both properties are `readOnly` denormalised mirrors, so **create/edit forms are unaffected** — `fieldsFromSchema` skips `readOnly` unless `includeReadOnly` is set, and `CnFormDialog` does not set it.

Three surfaces change:

1. **ClientDetail → Identity widget** — the intended fix. `CnObjectDataWidget` passes `includeReadOnly: true`, so readOnly is not a second gate.
2. **Clients index sidebar** — `CnIndexSidebar.allColumns` derives from the schema with no include, so Address and Website appear as newly toggleable columns. The table's own `config.columns` is explicit and unchanged.
3. **Client cards** — `CnObjectCard.metadataFields` filters only `visible === false` then slices to `maxMetadata`, so cards gain both fields and **other metadata may be pushed out**. `tests/e2e/visual/pipelinq.visual.spec.ts:23-24` (`clients.png`) may need re-baselining if the card view is the captured mode, as may the five `/clients` captures in `docs-screenshots.spec.ts`.

## ⚠️ Follow-up required on another branch

The assertion #775 describes under "Test coverage" is **not on `development`**. `tests/e2e/spec-coverage/declarative-view-system.spec.ts` lives on **`fix/gate-19-e2e-coverage`** (commit `30b1ef8e9`), where `:605-612` asserts `Address` and `Website` render **zero** times. That branch must flip the assertion before it merges, or it will fail against this change. Concretely:

- move `Address`/`Website` from the absence loop (`:605-612`) into the positive loop (`:598-604`);
- rewrite the #775 comment block (`:573-597`);
- **tighten the fixture picker at `:505-513`** — it currently requires `c.address && c.email && c.phone`; it needs `&& c.website`, because `demo_seed_data.json` seeds five clients all with an address but **only one with a website**, and the widget omits valueless fields. Without this the new assertion flakes exactly as the old comment describes.

## Class-wide check (#775's "worth checking alongside") — reported, not fixed

Resolved every `type: "data"` widget's `include` against the **merged** schema. Beyond `client-identity`, **7 more dead include entries across 3 widgets**, plus one phantom key — each is a separate product call about which side is wrong, so none are touched here:

| widget | dead entries |
|---|---|
| `client-commercial` | `notes` (`visible: false`) — renders 2 of 3 |
| `lead-deal` | `probability` (`visible: false`); `winProbability` is **not declared on any schema** |
| `lead-qualification` | `stageOrder`, `stageEnteredAt`, `description`, `notes` — 4 of 8 can never render |

The 9 data widgets with no `include` render all-visible properties and are unaffected. A `manifest include` ↔ `schema visible` cross-check would catch this whole class mechanically — `scripts/check-manifest.js` and `tests/validate-manifest.js` currently do not perform one, and `effective-manifest-crossref` already validates related invariants.

## Verification

- Real `deepMergeConfig()` before/after, with a positive control confirming the probe still finds 81 other `visible: false` properties (so "clean" means the flag is gone, not that the lookup missed).
- JSON parses; diff is two lines.
- gate-57 unaffected (no `lib/Service` change): 5 findings before and after, and a planted `post*` true positive was detected (6) and removed (5), proving the gate can still fail.

### DIFF
--- a/lib/Settings/register.d/15-unify-client-contact.json
+++ b/lib/Settings/register.d/15-unify-client-contact.json
@@ -43,7 +43,7 @@
             "type": "string",
             "title": "Address",
             "description": "Denormalised read-only mirror of the NC contact postal address. The Nextcloud Contact is authoritative.",
-            "visible": false,
+            "visible": true,
             "readOnly": true,
             "x-pipelinq-denormalised": true
           },
@@ -52,7 +52,7 @@
             "title": "Website",
             "description": "Denormalised read-only mirror of the NC contact website (vCard URL). The Nextcloud Contact is authoritative.",
             "format": "uri",
-            "visible": false,
+            "visible": true,
             "readOnly": true,
             "x-pipelinq-denormalised": true
           },


rubenvdlinde:
### Note: a green CI does **not** clear the visual-baseline risk

The E2E job runs the regression project only, and that project explicitly excludes the visual specs:

```
tests/e2e/playwright.config.ts:145
    testIgnore: ['**/docs-screenshots.spec.ts', '**/visual/**'],
```

So `tests/e2e/visual/pipelinq.visual.spec.ts-snapshots/clients-visual-linux.png` is never compared in CI. If the Clients surface is captured in card view, `CnObjectCard.metadataFields` will now include Address/Website and — because it slices to `maxMetadata` — may drop other fields, changing the image. That has to be checked with a local `npx playwright test --project visual` run; CI passing says nothing about it either way.

Same applies to the five `/clients` captures in `docs-screenshots.spec.ts` (`--project docs-capture`, also excluded from the regression project).

PR #176, opened 2026-03-25
feat: Add filters to Find Client widget

rubenvdlinde:
## Summary
- Add type (person/organization), address, and industry filter controls to the Find Client dashboard widget
- Replace simple text search with a full filter form — all filters combine with AND logic
- Industry dropdown is auto-populated from existing client data

Closes #175

## Test plan
- [ ] Open the dashboard and verify the Find Client widget shows the filter form
- [ ] Filter by type (person/organization) and verify results narrow correctly
- [ ] Type in the address filter and verify partial matching works
- [ ] Select an industry and verify only matching clients show
- [ ] Combine multiple filters with text search and verify AND logic

PR #779, opened 2026-08-11
feat(pipelinq): English vocabulary for the loyalty domain

rubenvdlinde:
Implements the loyalty slice of `openspec/changes/english-vocabulary`. Fifth app in the fleet programme, after openconnector #1213, opencatalogi #850, openbuild #176 and openregister #2421.

## Scope
43 renames in `70-loyalty-program.json` plus 22 consumer files. Beyond the ratified fleet words (`klant`→`customer`, `naam`→`name`, `bedrag`→`amount`, `aantal`→`quantity`, `prioriteit`→`priority`): `merk`→`brand`, `valuta`→`currency`, `taal`→`language`, `conditie`→`condition`, `formule`→`formula`, `drempel*`→`threshold*`, `*Beleid`→`*Policy`, `balansNa`→`balanceAfter`, `brondocument`→`sourceDocument`, `verwerktDoor`→`processedBy`, `beloning*`→`reward*`, `voorraad`→`stock`, `kanaal`→`channel`, and the schema `klantLoyaltyAccount`→`customerLoyaltyAccount`.

Validity dates take the ratified pair — `startdatum`/`geldigVan`→`validFrom`, `einddatum`/`geldigTot`→`validUntil` — verified first that no schema carries two words from either side.

## ⚠️ A runtime fatal that No syntax errors detected in Standard input code cannot see
`PosTransactionCompletedListener` called `processPosTransaction(klantId: ...)` as a **named argument** while the rename had already moved the parameter to `$customerId`. PHP resolves named arguments at call time: the linter passes, and it dies only when that listener fires. Four more of the same shape were in `PointsRuleEngineTest` (`formule:`, `conditie:`). Every named-argument call site is now checked against its declaration.

## Deliberately not renamed, each explained in the code
1. **The oc_appconfig key** `klantLoyaltyAccount_schema` — persisted state. Renaming it silently unconfigures existing instances; it needs a migration. Its *value* (the schema slug) did move.
2. **POS payload reads** — that entity isn't ours and may carry either vocabulary, so the `klantId`/`customerId`, `totaalbedrag`/`amount`, `kanaal`/`channel` fallbacks stay.
3. `zaakId` / `zaaktype` — held for the four-app procest window.

## Migration: 0 live objects, evidenced
21 shard tables exist for schema ids 492–498; all empty. Worth flagging how nearly this went wrong: my first count keyed on schema **title** and matched 1 of 7, because the registered title is *"Customer Loyalty Account"* while the slug is `klantLoyaltyAccount`. A positive control caught it — without one this would have recorded a confident, wrong zero.

### DIFF
--- a/l10n/en.json
+++ b/l10n/en.json
@@ -2571,7 +2571,7 @@
         "Creditcard": "Creditcard",
         "Currency": "Currency",
         "Customer": "Customer",
-        "Customer (klantId / contact UID)": "Customer (klantId / contact UID)",
+        "Customer (customerId / contact UID)": "Customer (customerId / contact UID)",
         "Customer portal": "Customer portal",
         "Customer-facing notes": "Customer-facing notes",
         "Date / time": "Date / time",
--- a/l10n/nl.json
+++ b/l10n/nl.json
@@ -2639,7 +2639,7 @@
         "Create loyalty account": "Loyaliteitsaccount aanmaken",
         "Credentials live on the OpenConnector source above, not on this provider row. Configure the vendor API key/secret on that source.": "Inloggegevens staan op de OpenConnector-bron hierboven, niet op deze aanbiedersregel. Configureer de API-sleutel/het geheim van de leverancier op die bron.",
         "Creditcard": "Creditcard",
-        "Customer (klantId / contact UID)": "Klant (klantId / contact-UID)",
+        "Customer (customerId / contact UID)": "Klant (customerId / contact-UID)",
         "Customer-facing notes": "Notities zichtbaar voor klant",
         "Date / time": "Datum / tijd",
         "Default country code (ISO-3166)": "Standaard landcode (ISO-3166)",
--- a/lib/BackgroundJob/PointsExpiryBatchJob.php
+++ b/lib/BackgroundJob/PointsExpiryBatchJob.php
@@ -181,21 +181,21 @@ class PointsExpiryBatchJob extends TimedJob
 
         if ($lastActivity < $noticeCutoff->format('c')) {
             // Advance notice.
-            $klantId = (string) ($account['klantId'] ?? '');
-            if ($klantId === '') {
+            $customerId = (string) ($account['customerId'] ?? '');
+            if ($customerId === '') {
                 return;
             }
 
             try {
                 $this->notificationService->sendNotification(
-                    userId: $klantId,
+                    userId: $customerId,
                     subject: 'loyalty_points_expiring',
                     parameters: [
                         'points'     => $balance,
                         'noticeDays' => $notice,
                         'accountId'  => $accountId,
                     ],
-                    objectType: 'klantLoyaltyAccount',
+                    objectType: 'customerLoyaltyAccount',
                     objectId: $accountId
                 );
             } catch (Throwable $e) {
--- a/lib/BackgroundJob/TierDowngradeJob.php
+++ b/lib/BackgroundJob/TierDowngradeJob.php
@@ -4,7 +4,7 @@
  * Pipelinq TierDowngradeJob.
  *
  * Daily job that processes scheduled tier downgrades. Iterates accounts whose
- * tierGeldigTot has passed, re-evaluates the tier via TierService, and applies
+ * tierValidUntil has passed, re-evaluates the tier via TierService, and applies
  * a downgrade when warranted (also emits the tier-changed event for notifications).
  *
  * @category BackgroundJob
@@ -99,8 +99,8 @@ class TierDowngradeJob extends TimedJob
 
             foreach ($accounts as $account) {
                 $accountId     = (string) $this->extractUuid(object: $account);
-                $tierGeldigTot = (string) ($account['tierGeldigTot'] ?? '');
-                if ($accountId === '' || $tierGeldigTot === '' || $tierGeldigTot > $now) {
+                $tierValidUntil = (string) ($account['tierValidUntil'] ?? '');
+                if ($accountId === '' || $tierValidUntil === '' || $tierValidUntil > $now) {
                     continue;
                 }
 
--- a/lib/BackgroundJob/TierDowngradeJob.php
+++ b/lib/BackgroundJob/TierDowngradeJob.php
@@ -98,7 +98,7 @@ class TierDowngradeJob extends TimedJob
             );
 
             foreach ($accounts as $account) {
-                $accountId     = (string) $this->extractUuid(object: $account);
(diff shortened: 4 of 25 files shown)


rubenvdlinde:
## Reviewed — three things to resolve before this can land

The rename itself is careful work: the named-argument fatal you caught is exactly the class the `RequireNamedParameters` sniff cannot see (it checks that a named argument is *present*, never that the name exists on the callee), and the title-vs-slug positive control on the migration count is the right instinct. Three findings, one of which is a live regression.

### 1. Blocker — the writer and the readers of the loyalty schema config key now disagree

The comment in `SchemaMapService` says the `oc_appconfig` key stays until a migration ships. Only the **readers** stayed. The **writer** moved.

`SettingsLoadService::SCHEMA_SLUGS` gained `customerLoyaltyAccount`, and `applySchemaConfig()` derives the key from it:

```php
foreach (self::SCHEMA_SLUGS as $slug) {
    ... setValueString(Application::APP_ID, "{$slug}_schema", ...);
}
```

so from this branch onward the only key ever written is **`customerLoyaltyAccount_schema`**. Meanwhile three call sites still read the old name:

| site | key read |
|---|---|
| `SettingsLoadService::applySchemaConfig()` **(writer)** | `customerLoyaltyAccount_schema` |
| `LoyaltyAccountService:410` | `customerLoyaltyAccount_schema` |
| `SchemaMapService:67` (`SCHEMA_MAPPING`) | `klantLoyaltyAccount_schema` |
| `LoyaltyReportingService:490` | `klantLoyaltyAccount_schema` |
| `SettingsService:114` (the settings whitelist `SchemaMapService` reads through) | `klantLoyaltyAccount_schema` |

Both failure modes are real and they are opposite:

- **Fresh install:** `klantLoyaltyAccount_schema` is never written, so `SchemaMapService::buildSchemaMap()` and `LoyaltyReportingService` resolve the loyalty-account schema to `''`.
- **Existing install:** `customerLoyaltyAccount_schema` does not exist until the register re-imports, so `LoyaltyAccountService` hits its own refusal path — and that path is fail-closed by design (*"register/customerLoyaltyAccount_schema not configured; OpenRegister calls are refused, not run unscoped"*). Loyalty account lookups stop working on upgrade until settings are reloaded.

Measured on the shared dev instance right now: `pipelinq.klantLoyaltyAccount_schema` = `495`, `pipelinq.customerLoyaltyAccount_schema` = unset. That is precisely the pre-upgrade state the second bullet describes.

The intent — keep the persisted key, move the slug — is right; it just needs the writer pinned to the old key too (an explicit slug→config-key map rather than string interpolation), or the migration written now. Whichever way it goes, one test should assert that the key `applySchemaConfig()` writes is the key `SchemaMapService` reads, because nothing in the suite currently connects those two ends.

### 2. The migration evidence is a live-object count, and the non-live set is not empty

`where _deleted is null` gives 0 across schemas 492–498 and I reproduce that. But `deleteObject()` is a soft delete, and counting every row:

| schema | slug | rows (all) | live |
|---|---|---|---|
| 495 | `klantLoyaltyAccount` | **6** | 0 |
| 496 | `pointsLedgerEntry` | **10** | 0 |
| 492, 493, 494, 497, 498 | — | 0 | 0 |
| 499, 500 | `giftCard`, `giftCardTransaction` | 0 | 0 |

All 16 are soft-deleted (dated 2026-06-21) and all carry the pre-rename property names. They are restorable, and after this lands they would come back with orphaned keys. Not a blocker on its own — but the sentence in the body should say *0 live objects on this instance*, not *0 objects*, and the same qualifier applies to any other install nobody has measured.

### 3. Direct conflict with #808, on lines whose content matters

Both PRs edit the same block of `PosTransactionCompletedListener::handle()`. #808 established, against the live schema, that `posTransaction` declares `customer` / `total` / `settledAt` / `reference` / `terminalId` and has **never** declared `klantId`, `totaalbedrag`, `voltooidOp`, `transactieId` or `kanaal` — so the read here resolves to `''` on every real transaction. This branch keeps only the three legacy keys:

```php
$customerId = (string) ($data['klantId'] ?? $data['customerId'] ?? $data['contactUid'] ?? '');
```

Whichever lands second must keep `$data['customer']` first. The comment above it — *"the POS transaction entity is not ours: it may carry either the Dutch or the English key"* — is a reasonable-sounding claim that the live schema contradicts, so it is worth restating in terms of what the schema actually declares.

### CI

The gates cell is red on gate-16 with 20 changed methods missing `@spec`, over a coverage line of 55 of 64 — so that is a real, wide scope, not a vacuous one. That needs closing on its merits rather than by narrowing anything.

Leaving this open. Happy to land it once the config-key split is closed; the rest of the rename reads correctly to me.

rubenvdlinde:
## Not merged in the fleet consolidation — superseded, and would undo two landed changes

Every other open PR touched in the last week was merged into `development` today (33 of them, several with `--admin` while red). This one and shillinq#495 were held back, because merging them would not produce red CI — it would silently undo work that is already live.

### The rename already landed, differently

`development` has merged its own English vocabulary for the loyalty domain, with migrations:

- 1c5f3e689 — translate Dutch vocabulary to English, with migration
- #826 — finish the translation, 20 names were half English half Dutch
- #831 — 20 pre-existing Dutch property names
- #837 — the stored enum values, and migrate them

Both sides are English. They just chose different words:

| this branch | `development` |
|---|---|
| `validFrom` / `validUntil` | `startDate` / `endDate` |
| `validUntil` | `validTo` |
| `tierValidUntil` | `tierValidTo` |
| `quantity` | `count` |
| `rewardValue` | `remunerationValue` |
| `brand` | `merk` |

`development`'s names are the ones its migrations have already moved stored objects to. Applying this branch's schema on top would declare properties the data is no longer under, and this branch carries no migration from `development`'s intermediate names — OpenRegister reads only what the schema declares, so those values go invisible without an error.

### It also reintroduces the stale container

`PointsRuleEngine` (and its siblings in this diff) still take:

```php
public function __construct(
    private ContainerInterface $container,
    ...
```

That is the pre-ADR-083 lazy-lookup pattern. It was removed across 77 sites in this app and four others and replaced with an injected `ObjectServiceInterface` (ADR-084, openregister#2498). Merging this branch puts it back in pipelinq.

### What this needs

A rebase, not a conflict resolution: keep `development`'s chosen names, and bring across only what this branch adds that `development` lacks — the frontend/portal/test follow-through in f8338f8ca and the l10n and phpcs alignment in 4f834889e — retargeted at `development`'s vocabulary and its injected constructor.

PR #780, opened 2026-08-11
fix(booking): defer the walk-in queue rebalance to its background job

rubenvdlinde:
## What

`BookingService::completeBooking()` called `WalkInQueueService::rebalance()` **synchronously**, inside the HTTP request that completes a booking. This PR defers it to `WalkInQueueRebalanceJob`, which already existed and which the spec already names.

## The measurement, not an estimate

`rebalance()` reads every waiting walk-in ticket (capped at `QUEUE_PAGE_SIZE = 200`) and, per ticket, does a service read, a resource read, an availability computation and a `saveObject()`.

A new test seeds a full queue and pins the fan-out:

| | measured |
|---|---|
| `saveObject()` calls per `rebalance()` at full queue | **200** |
| availability computations per `rebalance()` at full queue | **200** |

`WalkInQueueServiceTest::testRebalanceFansOutOneSaveObjectPerWaitingTicket`.

All of that sits on the request path, and **every error was swallowed** at the seam (`BookingService.php:718` on `206db4be`), so a total failure of the rebalance was invisible to the caller while request latency scaled with queue depth. This is the openregister#2420 family — the same shape that serialised every object write on a live instance for over two hours on 11 August.

### 🔴 Correction to my own first draft — the fan-out is LATENT, not live

I originally wrote that the 200 writes "ran on the request path". **That overstates what I measured.**

`WalkInQueueService::listByStatus()` (`:556-562`) and `fetchResourceRows()` (`:727-733`) supply `register`/`schema` at the **top level** of the `findAll()` config. The real OpenRegister reads that context **only** from `$config['filters']` — `ObjectService::prepareFindAllConfig()`, verified against `openregister@a4dd9067` — and `MagicMapper::findAll()` (`:8681`) then returns `[]` with a warning log. So the ticket read currently comes back empty.

| | |
|---|---|
| what `rebalance()` does per waiting ticket, at `QUEUE_PAGE_SIZE` | **200 `saveObject()` + 200 availability computations** — measured |
| what a booking completion writes **today** | **1** (the booking), because the queue read is mis-keyed |
| what it writes **the moment that mis-keying is fixed**, without this PR | **201** |

The number came from **a double I wrote**, which keys off the top level and therefore does not reproduce the upstream contract — it measures the loop, not production. The test docblock now says so (`3386729de`).

**This makes the PR more load-bearing, not less.** The mis-keying is a separate defect at **11 call sites** (pipelinq#793, root-caused upstream as openregister#2426). Whoever fixes it converts a latent 201-write request into a live one — **so this must land first, or that fix ships a performance incident with it.**

⚠️ Worth stating explicitly: **gate-61 structurally cannot see this.** It inspects post-event *listeners*; this is a *service seam* invoked from a controller. A clean gate-61 says nothing about it. (And gate-61 is dead on every `--full` run anyway — `.github#347`.)

## No spec was authored for this

`openspec/specs/appointment-booking/spec.md`, scenario *"Queue rebalances as appointments complete"*, already reads:

> **THEN** the **WalkInQueueRebalanceJob** MUST recalculate `estimatedReadyAt` for all waiting tickets

and `WalkInQueueRebalanceJob`'s own docblock already said it runs *"only when explicitly scheduled by `BookingService::completeBooking` via the JobList"*. **Nothing ever scheduled it.** The deferred path was built and never switched on, and the synchronous path was carrying the load. This PR moves the code onto a requirement that was already written; no spec file is touched.

## Changes

- `BookingService` takes `IJobList` and enqueues `WalkInQueueRebalanceJob` instead of running the fan-out inline. `IJobList::add()` is idempotent for an identical `(class, argument)` pair, so N completions in one cron window collapse to **one** rebalance — which is also strictly more correct, since the rebalance is a whole-queue recomputation and running it once after the last completion produces the same ETAs as running it after each.
- `WalkInQueueRebalanceJob` now extends `QueuedJob` rather than `Job`, so an enqueued instance runs once and removes itself from the job list. A plain `Job` on the list would re-run on every cron tick.
  - It is deliberately **not** added as a `<job>` in `appinfo/info.xml`. That channel registers **periodic** work; this job is on-demand. "The job exists but is unregistered" is true, but `info.xml` is the wrong registration for it.
- The `setWalkInQueueRebalance()` seam and its bootstrap wiring in `Application::bootNonPageSurfaces()` are **removed**, not kept alongside the job. That wiring also resolved `BookingService` **and** `WalkInQueueService` out of the container on every non-page request.

`WalkInQueueRebalanceJob::run()` is now the only caller of `rebalance()` anywhere in `lib/` or `src/` — verified with a positive control on the same grep, because "no callers" is an absence claim.

## Verification

| check | result |
|---|---|
| PHPUnit `Unit Tests` | **1639 tests, 5150 assertions, 0 failures** |
| phpcs | clean on all touched files |
| phpmd | clean |
| phpstan | `No errors` |
| psalm | `No errors found!` |

Also fixed **seven pre-existing phpcs errors** in `BookingServiceTest.php` encountered on the way — one of which was a docblock that had become **detached** from `testCompleteBookingRejectsTransitionFromTerminalStatus` and was sitting in front of an unrelated test, leaving that test undocumented.

## What this PR does not claim

The write-budget test (`testCompleteBookingWritesOnlyTheBookingItself`) constrains a re-introduced fan-out that goes through the container-resolved `ObjectService`. It **cannot** see one routed through a separately-injected collaborator — which is why the enqueue itself is asserted separately rather than inferred from the write count.

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -67,7 +67,6 @@ use OCA\Pipelinq\Service\BsnValidationService;
 use OCA\Pipelinq\Service\Gdpr\PipelinqApRegulatorEscalateProvider;
 use OCA\Pipelinq\Service\Gdpr\PipelinqBsnIdentityVerifyProvider;
 use OCA\Pipelinq\Service\HaalCentraalClient;
-use OCA\Pipelinq\Service\WalkInQueueService;
 use Throwable;
 use OCP\App\IAppManager;
 use OCP\AppFramework\App;
@@ -805,7 +804,6 @@ class Application extends App implements IBootstrap
     {
         $this->registerCommentResolvers(server: $server);
         $this->wireAppointmentEmailSeam();
-        $this->wireBookingWalkInRebalance();
         $this->wireAppointmentCalendarSeam();
         $this->wireAppointmentPaymentSeam();
         $this->wireGdprSeamProviders();
@@ -843,24 +841,6 @@ class Application extends App implements IBootstrap
         }//end try
 
     }//end registerCommentResolvers()
-
-    /**
-     * Wire the walk-in queue rebalance seam into the booking lifecycle, so a
-     * Booking completion fires WalkInQueueService::rebalance.
-     *
-     * @return void
-     */
-    private function wireBookingWalkInRebalance(): void
-    {
-        try {
-            $bookingService     = $this->getContainer()->get(BookingService::class);
-            $walkInQueueService = $this->getContainer()->get(WalkInQueueService::class);
-            $bookingService->setWalkInQueueRebalance(service: $walkInQueueService);
-        } catch (\Exception $e) {
-            // Booking / walk-in surfaces not available — leave rebalance seam unset.
-        }
-
-    }//end wireBookingWalkInRebalance()
 
     /**
      * Read the SPA manifest's declared app dependencies, as a flat id list.
--- a/lib/BackgroundJob/WalkInQueueRebalanceJob.php
+++ b/lib/BackgroundJob/WalkInQueueRebalanceJob.php
@@ -30,20 +30,25 @@ namespace OCA\Pipelinq\BackgroundJob;
 
 use OCA\Pipelinq\Service\WalkInQueueService;
 use OCP\AppFramework\Utility\ITimeFactory;
-use OCP\BackgroundJob\Job;
+use OCP\BackgroundJob\QueuedJob;
 use Psr\Log\LoggerInterface;
 
 /**
  * On-demand background job that rebalances walk-in ticket ETAs.
  *
- * Extends `OCP\BackgroundJob\Job` (NOT TimedJob) — runs only when explicitly
- * scheduled by BookingService::completeBooking via the JobList. Each invocation
- * delegates to {@see WalkInQueueService::rebalance()} and logs the touched
- * ticket count.
+ * Extends `OCP\BackgroundJob\QueuedJob` (NOT TimedJob, and NOT a plain Job) —
+ * it runs only when explicitly scheduled by BookingService::completeBooking via
+ * `IJobList::add()`, and removes itself from the job list once it has run. A
+ * plain `Job` added to the list would re-run on every cron tick forever, which
+ * is the opposite of "on-demand"; this job is deliberately never listed as a
+ * `<job>` in appinfo/info.xml, because that channel registers PERIODIC work.
+ *
+ * Each invocation delegates to {@see WalkInQueueService::rebalance()} and logs
+ * the touched ticket count.
  *
  * @spec openspec/specs/appointment-booking/spec.md
  */
-class WalkInQueueRebalanceJob extends Job
+class WalkInQueueRebalanceJob extends QueuedJob
 {
     /**
      * Constructor.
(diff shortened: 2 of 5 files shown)


rubenvdlinde:
## 🔴 Correction to this PR's own headline claim — pushed as `3386729de`

I wrote that the 200 `saveObject()` calls "ran on the request path". **That overstates what I measured, and I am correcting it before this merges.**

`WalkInQueueService::listByStatus()` (`:556-562`) and `fetchResourceRows()` (`:727-733`) supply `register`/`schema` at the **TOP LEVEL** of the `findAll()` config. The real OpenRegister reads that context **only** from `$config['filters']` — `ObjectService::prepareFindAllConfig()`, verified against `openregister@a4dd9067` — and `findAll()` then passes `$this->currentRegister` / `$this->currentSchema` to the handler. With no context resolved, `MagicMapper::findAll()` (`:8681`) **returns `[]` with a warning log**.

So on a live instance the ticket read currently comes back empty and the fan-out is **LATENT, not live**. The `ObjectService` double in my test keys off the top level — it does not reproduce the upstream contract — so it measures **what the loop performs per waiting ticket**, which is the honest reading of the number.

### The corrected claim

| | |
|---|---|
| what `rebalance()` does per waiting ticket, at `QUEUE_PAGE_SIZE` | **200 `saveObject()` + 200 availability computations** — measured |
| what a booking completion writes **today** | **1** (the booking), because the queue read is mis-keyed |
| what it would write **the moment that mis-keying is fixed**, without this PR | **201** |

### Why this makes the PR more load-bearing, not less

The mis-keying is a separate defect and it is not a one-off: the same shape appears at **11 call sites** across `lib/` (filed separately). Whoever fixes it converts a latent 201-write request into a live one. **This PR must land first, or that fix ships a performance incident with it.**

Nothing else in the PR changes: the spec already mandated the deferred job, `WalkInQueueRebalanceJob` already existed and was never scheduled, and `WalkInQueueRebalanceJob::run()` remains the only caller of `rebalance()` anywhere in `lib/` or `src/`.

🔑 Recording the method that caught this, because it nearly did not: the number came from **my own double**, and a double I wrote to exercise a loop is not evidence about production. The contradiction only surfaced when a second, independent reader checked where OpenRegister actually reads register/schema. **When a measurement comes from a test harness you authored, name the layer it covers.**

rubenvdlinde:
⚠️ **Merge order:** this PR must land **before** #796. #796's  calls `new BookingService(...)` without `jobList` and calls `setWalkInQueueRebalance()` — both of which this PR changes/removes. Neither PR's own CI can see the collision, because each is green against `origin/development` as it stands today. Details and the follow-up edit are described in ConductionNL/pipelinq#796.

rubenvdlinde:
⚠️ **Correcting my previous comment — a backticked filename was eaten by shell substitution before it was posted.** The full version:

**Merge order: this PR must land BEFORE #796.**

#796's `tests/Unit/Controller/BookingAdminControllerTest.php` calls `new BookingService(...)` **without** `jobList` and calls `setWalkInQueueRebalance()` — both of which this PR changes or removes. Merging in either order without an edit lands a red `development`.

Neither PR's own CI can see the collision, because each is green against `origin/development` as it stands today. Details and the follow-up edit are in #796.

Issue #795, opened 2026-08-11
GET /api/rapportage/export answers 501 while generateCsv()/neutralizeCsvCell() exist and are correct but unreachable from the route


rubenvdlinde:
## `GET /api/rapportage/export` is a stub that answers 501, while the machinery behind it exists and works

`lib/Controller/ReportingController.php:260-273` returns **501 "Export not yet implemented"** for every authenticated caller — while its own docblock carries `@spec …#requirement-export-and-bi-integration`.

Meanwhile `ReportingService::generateCsv()` and `ReportingService::neutralizeCsvCell()` already exist and are **correct**: a leading `'` is prefixed to cells beginning `=`, `+`, `-`, `@`, tab or CR; cells containing a comma or quote are quote-wrapped; embedded `"` is doubled. They are reachable only from `ForecastExportService`.

So the CSV-injection contract is implemented, tested and unreachable from the route that is supposed to expose it.

- `ReportingControllerTest::testExportCsvReturnsANeutralisedCsvDownload` asserts the intended contract and is `markTestSkipped('BUG: …')` naming file:line.
- `testCsvWriterNeutralisesInjectionPrefixesAndQuotes` pins the escaping contract where it *is* currently reachable, and **passes**.

Either wire the route to `generateCsv()`, or drop the `@spec` tag and the route — but a routed endpoint whose docblock claims a requirement it answers 501 for is the shape that makes a spec-coverage number meaningless.

⚠️ Note for whoever wires it: `DataDownloadResponse::getHeaders()` cannot be asserted in the bare unit environment — `OCP\AppFramework\Http\Response:232` reaches for the `OC` class. Assert `render()` and the input to `toCsv()` instead.

Found while writing gate-25 contract tests.

Issue #794, opened 2026-08-11
ObjectService::find() called positionally with register/schema in the $_extend/$files slots — contract transition 404s and Customer 360 denies every read


rubenvdlinde:
## Summary

Two call sites pass `register` and `schema` **positionally** as the 2nd and 3rd arguments to OpenRegister's `ObjectService::find()`. The real signature puts `$_extend` and `$files` there:

`openregister@a4dd9067`, `lib/Service/ObjectService.php:635-644`:

```php
public function find(
    int|string $id,
    ?array $_extend = [],                       // <-- 2nd
    bool $files = false,                        // <-- 3rd
    Register|string|int|null $register = null,  // <-- 4th
    Schema|string|int|null $schema = null,      // <-- 5th
    ...
): ?ObjectEntity
```

A `string` into `?array $_extend` is a `TypeError` under **any** coercion mode — `string → array` is never coercible — and both call sites are in files carrying `declare(strict_types=1)`. Both are wrapped in `catch (Throwable)`, so the TypeError is swallowed and turned into "not found".

### 1. `lib/Controller/ContractController.php:248` — every contract 404s

```php
$object = $objectService->find($id, $registerId, $schemaId);
```

`loadContract()` returns `null` for every existing contract, so **`POST /api/contracts/{id}/transition` is dead** — it answers 404 for every id. Pinned by `ContractControllerTest::testTransitionResolvesTheContractAgainstTheUpstreamFindSignature` (`markTestSkipped`, message names the bug).

### 2. `lib/Controller/Customer360Controller.php:142` — Customer 360 denies every read

```php
$object = $objectService->find($clientId, $register, $schema);
```

This is `canReadClient()`, the per-object read guard. It **fails closed** by design — the docblock says so explicitly:

> *Fails closed — an OR outage or missing config denies the read rather than granting it, since this is the caller's only defense against reading another client's data.*

So this is **not** a security hole; it is the safe direction. But it means the guard denies **every** caller and the Customer 360 surface is entirely non-functional.

## ⚠️ A test would have gone green over this

`tests/Stubs/Service/ObjectService.php:47` declares:

```php
find(string $id, string $register = '', string $schema = '')
```

— i.e. **the wrong signature**, matching the buggy call rather than the real one. Any test using the bundled stub passes over this bug. The pinning test above deliberately declares the upstream signature instead.

[...]

Issue #793, opened 2026-08-11
11 findAll() call sites key register/schema at the top level, where OpenRegister never reads them — permanently empty results behind a healthy 200


rubenvdlinde:
## Summary

Eleven `findAll()` call sites supply `register` / `schema` at the **top level** of the config array instead of inside `filters`. OpenRegister never reads them there, so the query resolves **no register/schema context** and `MagicMapper::findAll()` returns `[]` — with a warning log, no exception. The endpoint above it answers a healthy **HTTP 200 with an empty result**, permanently.

This is the decidesk shape, replicated eleven times.

## The mechanism, verified against the OpenRegister source

`openregister@a4dd9067`, `lib/Service/ObjectService.php`:

```php
private function prepareFindAllConfig(array $config): array
{
    ...
    if (isset($config['filters']['register']) === true && ...) {   // <-- filters ONLY
        $this->setRegister(register: $config['filters']['register']);
    }
    if (isset($config['filters']['schema']) === true && ...) {     // <-- filters ONLY
        $this->setSchema(schema: $config['filters']['schema']);
    }
    return $config;
}
```

`findAll()` then hands the handler `register: $this->currentRegister, schema: $this->currentSchema`, and `MagicMapper::findAll()` (`lib/Db/MagicMapper.php:8681`) begins:

```php
if ($register === null || $schema === null) {
    $this->logger->warning('[MagicMapper] findAll() called without register/schema context', ...);
    return [];
}
```

⚠️ **Worse than "always empty": it is non-deterministic.** `$currentRegister` / `$currentSchema` are **shared instance state** on `ObjectService`. `find()` was explicitly hardened to snapshot-and-restore them (BUG-OBJ-13 / openregister#1520, after `TransitionController` 500-ed on an inherited stale schema) — but `findAll()` does not restore. So a mis-keyed call gets **whatever register/schema the last correctly-keyed call in the same request happened to leave behind**, and only returns `[]` if there was none. The result therefore depends on call order within a request.

## The eleven

| file:line | what it reads |
|---|---|
| `lib/Service/SlaAttainmentService.php:456` | SLA breach events |
| `lib/Service/SlaAttainmentService.php:631` | SLA breach events |
| `lib/Service/SlaEngineService.php:565` | SLA engine |
| `lib/Service/WalkInQueueService.php:556` | **every waiting walk-in ticket** |
| `lib/Service/WalkInQueueService.php:727` | **candidate resources for the ETA** |
[...]

Issue #792, opened 2026-08-11
contactSync writeBack answers success:true over a no-op and can create 'Unknown' vCards; import is not idempotent


rubenvdlinde:
## 🟠 `contactSync#writeBack` reports `success: true` over a no-op — and can create junk contacts

`lib/Controller/ContactSyncController.php:220-225` returns `200 {success: true, contactsUid: null}` whenever `syncToContacts()` returns `null` — which covers *Contacts app unavailable*, *object not found* **and** *vCard write failed*. Measured.

One level down it gets worse. `lib/Service/ContactVcardService.php:230-260` declares `fetchPipelinqObject(): ?array` but returns `serializeResult($object)`, and `serializeResult()` (`:269-280`) returns **`[]` for `null`**. So the `if ($objData === null)` guard at `:93` **never fires** for a missing or inaccessible object. Execution continues into `buildProperties([], 'client')`, which yields `['FN' => 'Unknown']` (`ContactVcardPropertyBuilder.php:59`), and `writeToAddressBook()` creates a vCard named **"Unknown"** in the caller's default addressbook.

So `POST /api/contacts-sync/write-back` with a bogus `objectId` creates junk contacts and answers `success: true`. The controller half is measured; the `[]`-for-`null` chain is established by reading `serializeResult`.

🔑 This is the `?array` return type lying: a nullable-returning method that funnels through a serialiser which maps `null → []` can never return `null`, so every `=== null` guard downstream is dead. Worth grepping for.

## 🟠 `contactSync#import` is not idempotent

`lib/Service/ContactImportService.php:130-136` always calls `saveObject(…, null)`. Nothing looks for an existing object carrying the same `contactsUid`, and neither the `client` nor the `contact` schema constrains it (`lib/Settings/pipelinq_register.json`).

**Measured:** two imports of one addressbook UID produced two creates. One person forks into two records that then drift independently.

Test: `ContactSyncControllerTest::testImportSameContactTwiceIsIdempotent`, `markTestSkipped('BUG: …')`.

## What came back CLEAN

`contactSync#import`'s **field survival** is fine — both a sparse and a rich contact keep every unrelated field. The PUT-semantic erasure shape was probed for specifically and does not occur on this path.

## Test-infrastructure blockers found on the way (not product bugs, but they cost time)

[...]

PR #796, opened 2026-08-11
test(contract): close 96 gate-25 findings with real contract tests — 13 product defects found

rubenvdlinde:
## What

gate-25 `contract-coverage` was pipelinq's largest gate at full scope: **96** routed, publicly-reachable controller methods with no automated proof of their wire contract. This closes all 96.

```
canonical ConductionNL/.github@765d081 · env -u HYDRA_GATE_BASE_REF · tree 206db4be
before:  [gate-25] contract-coverage: FAIL — 96 new public endpoint(s) without a contract test
after:   [gate-25] contract-coverage: PASS — 228 new endpoint(s), all covered
```

⚠️ **Name the base.** 96 is the **full-scope** figure on `origin/development`. The beta-diff table read **2** — the narrow scope understated this by 48×.

**488 new tests across 34 files.** Suite: **2127 tests / 7019 assertions / 0 failures** (base 1639 / 5150). No change to `lib/`, `src/`, `appinfo/`, `composer.json` or `phpunit.xml`.

## 🔴 The point was never the number — 13 product defects

Every one of them was answering a **healthy 200/201**. Eight were proven by running the correct-contract assertion, observing the failure, and only then marking it.

| # | defect |
|---|---|
| **#781** | **IDOR ×4 on the loyalty surface.** `POST /api/loyalty/redemption/initiate/{accountId}/{optionId}` returns **201**, debits the victim's balance, and hands the **caller** the reward code. `getAccount`'s own docblock says *"the caller MUST own the underlying klantId"* |
| **#782** | `mfaEnroll` writes `mfaEnabled = false` before any verification — a hijacked portal session strips MFA. Measured end-to-end: `mfa-required` → `authenticated` |
| **#783** | POS consent sync saves a **10-key display projection** over the contact; `saveObject` is PUT-semantic, so every other field is nulled |
| **#784** | POS payment webhook answers **200** for a delivery it never processed — the provider never retries, the payment is lost |
| **#785** | gift-card redeem is not idempotent (a POS retry debits twice); activate returns 200 for a **blocked** card |
| **#786** | `resolvePrice` prices the **client's own request body**, against two docblocks saying it does not |
| **#787** | `/segments/{id}/size` returns and **persists** a stale cached count, and swallows a failed write behind a 200 |
| **#788** | destination credential filter is a denylist missing `token` / `apiKey` / `connectionString` |
| **#789** | the **AVG Art. 15** data export delivers no data — `buildExport()` has **zero callers** |
| **#790** | 🔴 **CTI webhook verifies the signature ONLY IF ONE IS PRESENT** — an unsigned POST writes a contactmoment on a `#[PublicPage]` route |
| **#791** | BRP address-reveal audit entries hash the **empty string**, so every entry identifies the same nobody; a failed audit write does not stop the reveal |
| **#792** | `contactSync` writeBack answers `success: true` over a no-op and can create "Unknown" vCards |
| **#793** | 🔴 **11 `findAll()` call sites key `register`/`schema` at the top level**, where OpenRegister never reads them — permanently empty results behind a healthy 200 |
| **#794** | `ObjectService::find()` called **positionally** with register/schema in the `$_extend`/`$files` slots — contract transition 404s for every id, Customer 360 denies every read |
| **#795** | `GET /api/rapportage/export` answers 501 while its CSV writer exists, is correct, and is unreachable from the route |

Also filed upstream: **openregister#2426** (`findAll()` documents two config keys it never reads — the root cause of #793) and **.github#365** / **#367**.

## ⚠️ This PR's OWN gate-25 cell will read EMPTY SCOPE — that is correct, and it is not the evidence

The PR run is diff-scoped (`HYDRA_GATE_BASE_REF=origin/development`) and this PR touches **no** `lib/Controller/` file, so the gate filters its scope to nothing. Measured on this exact branch:

```
[gate-25] contract-coverage: EMPTY SCOPE — diff-scoped against 'origin/development' and NO
controller file was touched, so no endpoint was inspected. Wire-contract coverage is
UNVERIFIED by this run. This is not a pass.
```

Credit to the checker for saying *"This is not a pass"* out loud — most gates in the suite do not.

**So do not read this PR's gate-25 cell as the result.** The evidence for 96 → 0 is the full-scope run quoted at the top, reproducible with the exact command given. This is the inverse of the gate-19 remediation case: a gate-19 fix necessarily edits spec files and therefore *does* enter its own scope, whereas a contract-test PR never touches a controller and therefore never enters gate-25's.

## How the tests were written, and what they are worth

Each endpoint gets a test that constructs the **real** controller, calls the method, and asserts the **status code** *and* the **response body shape** — never a bare call.

Where the contract lives *below* the controller — price arithmetic, read-modify-write field survival, IDOR guards, HMAC verification, aggregate correctness — the **real service** is wired over an in-memory `ObjectService` double, because a mocked facade cannot demonstrate a guard. The doubles reproduce the upstream contract deliberately: register/schema read **only** from `$config['filters']`, empty result when no context resolves, reserved filter keys excluded, soft-deleted rows excluded, `limit`/`offset` honoured. **That fidelity is what found #793.**

The aggregate endpoints assert **seeded values**, not "status is 200". That distinction is the entire reason #793 and the SLA-attainment defect surfaced.

## The honesty checks, because this gate is weak

`check_contract_coverage.py`'s PHPUnit test is a bare substring search for `->methodName(` over the **concatenated** text of every `tests/**/*Test.php`. It is not scoped to the class, counts no assertions, and **cannot see `markTestSkipped`**. So:

- **No `@contract exclude` was added anywhere.**
- **No endpoint is credited only by a skipped test.** Verified mechanically: splitting every test file into methods, discarding those containing `markTestSkipped`, and searching only the remainder finds a live `->method(` call for **all 96** — `0` endpoints credited by skipped text alone.
- 27 tests are committed as `markTestSkipped('BUG: …')` so they assert the **correct** contract and go green the moment it holds. Each message names the file:line of the defect.

⚠️ **Measured false coverage that already existed before this PR** (filed as .github#367): `posCustomer#search` was credited by `->search(` in `XWikiControllerTest`/`KvkApiClientTest`; `posPayment#initiate` and `#refund` by calls on **payment adapters**. None of those three had a single test. They do now.

⚠️ **gate-25 keys on `#[PublicPage]`/`#[NoAdminRequired]`**, so admin-only endpoints are invisible to it — e.g. `cti#getConfig`/`updateConfig`/`testConnection`/`eventLog`, which dropped the attribute deliberately. **"gate-25 green" says nothing about admin endpoints.**

## Two test-infrastructure defects found on the way

- `tests/Stubs/Service/ObjectService.php:47` declares `find(string $id, string $register='', string $schema='')` — **the wrong signature**, matching the buggy caller in #794 rather than the real callee. Any test using the stub goes green over that bug. Its docblock claims it *"mirrors the real OR ObjectService signature"*; for `findAll`/`count` it does, for `find` it does not. Not fixed here — fixing it should be its own PR, with the suite re-run to see what else it was hiding.
- `MessagingWebhookController::readRawBody()` is `private`, whereas the house pattern (`BlastWebhookController::readRawBody()`, explicitly documented as *"indirected so tests can override"*) is `protected`; `CtiController::webhook()` and `BrpController::mutationWebhook()` inline `file_get_contents('php://input')` with no seam at all.

### DIFF
--- a/tests/Unit/Controller/ActivityTimelineControllerTest.php
+++ b/tests/Unit/Controller/ActivityTimelineControllerTest.php
@@ -1,0 +1,422 @@
+<?php
+
+/**
+ * Contract tests for ActivityTimelineController.
+ *
+ * Covers `GET /api/timeline`, `GET /api/worklog` and `POST /api/worklog` —
+ * status codes, response body shape, the required-parameter rejections and the
+ * entity-existence probe.
+ *
+ * @category Test
+ * @package  OCA\Pipelinq\Tests\Unit\Controller
+ *
+ * @author    Conduction <info@conduction.nl>
+ * @copyright 2026 Conduction B.V.
+ * @license   EUPL-1.2 https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12
+ *
+ * @version GIT: <git_id>
+ *
+ * @link https://pipelinq.nl
+ *
+ * SPDX-FileCopyrightText: 2026 Conduction B.V. <info@conduction.nl>
+ * SPDX-License-Identifier: EUPL-1.2
+ */
+
+declare(strict_types=1);
+
+namespace OCA\Pipelinq\Tests\Unit\Controller;
+
+use OCA\Pipelinq\Controller\ActivityTimelineController;
+use OCA\Pipelinq\Service\ActivityTimelineService;
+use OCP\AppFramework\Http;
+use OCP\IRequest;
+use OCP\IUser;
+use OCP\IUserSession;
+use PHPUnit\Framework\TestCase;
+use Psr\Container\ContainerInterface;
+use Psr\Log\LoggerInterface;
+
+/**
+ * ActivityTimelineController contract coverage.
+ *
+ * @SuppressWarnings(PHPMD.TooManyPublicMethods) Happy path plus at least one
+ *  contract-relevant failure path for each of the three endpoints.
+ */
+class ActivityTimelineControllerTest extends TestCase
+{
+    /**
+     * The request double.
+     *
+     * @var IRequest
+     */
+    private IRequest $request;
+
+    /**
+     * The timeline service double.
+     *
+     * @var ActivityTimelineService
+     */
+    private ActivityTimelineService $service;
+
+    /**
+     * The user session double.
+     *
+     * @var IUserSession
+     */
+    private IUserSession $userSession;
+
+    /**
+     * In-memory OpenRegister ObjectService double used by the existence probe.
+     *
+     * @var object
+     */
+    private object $objects;
+
+    /**
+     * Set up the doubles.
+     *
+     * @return void
+     */
+    protected function setUp(): void
+    {
+        $this->request     = $this->createMock(IRequest::class);
+        $this->service     = $this->createMock(ActivityTimelineService::class);
+        $this->userSession = $this->createMock(IUserSession::class);
+
+        $this->objects = new class {
+            /**
+             * Rows keyed by id.
+             *
+             * @var array<string, array<string, mixed>>
+             */
+            public array $store = [];
+
+            /**
+             * Read one row.
+             *
+             * @param int|string $id      Object id.
+             * @param array|null $_extend Extend list.
+             * @param bool       $files   Include files.
+             *
+             * @return array<string, mixed>|null
+             */
+            public function find(int|string $id, ?array $_extend=[], bool $files=false): ?array
+            {
+                return ($this->store[(string) $id] ?? null);
+            }//end find()
+        };
+    }//end setUp()
+
+    /**
+     * Build the controller under test.
+     *
+     * @return ActivityTimelineController
+     */
+    private function buildController(): ActivityTimelineController
+    {
+        $container = $this->createMock(ContainerInterface::class);
+        $container->method('get')->willReturnCallback(
+            function (string $id): object {
+                if ($id === 'OCA\\OpenRegister\\Service\\ObjectService') {
+                    return $this->objects;
+                }
+
+                throw new \RuntimeException('not registered: '.$id);
+            }
+        );
+
+        return new ActivityTimelineController(
+            request: $this->request,
+            service: $this->service,
+            userSession: $this->userSession,
+            logger: $this->createMock(LoggerInterface::class),
+            container: $container,
+        );
+    }//end buildController()
+
+    /**
+     * Sign a user in.
+     *
+     * @return void
+     */
+    private function signIn(): void
+    {
+        $user = $this->createMock(IUser::class);
+        $user->method('getUID')->willReturn('agent-1');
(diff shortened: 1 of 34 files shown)


rubenvdlinde:
## ⚠️ Merge order matters between #780 and #796 — they collide

**#780 must merge first, and #796 must then be rebased and updated before it merges.**

#780 changes `BookingService`'s constructor (adds `IJobList`) and **removes** `setWalkInQueueRebalance()`. #796's `tests/Unit/Controller/BookingAdminControllerTest.php` does both of the things that removes:

```php
// setUp(), :158-169
$this->bookings = new BookingService( container: …, appConfig: …, userSession: …,
    availabilityService: …, eligibilityService: …, logger: … );   // no jobList
$this->bookings->setWalkInQueueRebalance(service: $this->walkIn);  // removed by #780
```

Merging them in either order without an edit lands a **red `development`**. Neither PR's own CI can see this — each is green against `origin/development` as it stands today, which is exactly why it needs saying out loud.

### What has to change in #796 after #780 lands

Three tests, and they get **stronger**, not weaker:

| test | after #780 |
|---|---|
| `testMarkCompletedIssuesOnlyTheBookingWriteWhileTheQueueReadIsMisKeyed` | still 1 write, but now for a *different and better* reason — the rebalance is deferred, not merely starved by the mis-keyed query. Reframe the docblock. |
| `testMarkCompletedFansOutOneWritePerWaitingTicketInsideTheRequest` | **inverts into the regression guard for #780**: with `acceptTopLevelContext = true` — i.e. the queue query working — the request must *still* issue exactly **one** write and enqueue `WalkInQueueRebalanceJob`. That is a much stronger assertion than the one it replaces, because it proves the deferral holds under the conditions that would otherwise produce 201 writes. |
| `testMarkCompletedStillReportsSuccessWhenTheQueueRebalanceThrows` | the seam is gone; becomes "still reports success when `IJobList::add()` throws". |

I will do this rebase myself once #780 is merged.

rubenvdlinde:
✅ **Sequencing done.** #780 merged as `67cb89b89`; `origin/development` merged into this branch and the collision resolved in `ea1c8018c`.

The collision was **measured, not predicted** — merging the two produced **10 errors**, the whole of `BookingAdminControllerTest`.

Two of the three affected tests came out stronger. `testMarkCompletedFansOutOneWritePerWaitingTicketInsideTheRequest` is now `testMarkCompletedDefersTheQueueRebalanceEvenWithTheQueryKeysCorrected`: it holds `acceptTopLevelContext = true`, i.e. keeps the queue read **working** — the world after #793 is repaired — and demands the request still issue exactly **one** write and enqueue the job. Positive-controlled with a temporary seam restoring the pre-#780 synchronous call: it fails with `Failed asserting that actual size 201 matches expected size 1`. The guard can fail, and at exactly the right number.

Suite: **2130 tests, 7225 assertions, 0 failures**. gate-25 unchanged at full scope: `PASS — 228 new endpoint(s), all covered`.

PR #797, opened 2026-08-11
fix(navi): mint the conversationId and answer follow-ups in context (#778)

### LINKED ISSUE #778: Navi conversationId is never minted: the controller hardcodes null and only echoes the client, so `conversational-follow-up` has no server half
`openspec/specs/dashboard/spec.md#conversational-follow-up` requires two things:

> - THEN the frontend MUST include the `conversationId` in subsequent requests to maintain context
> - AND **the response MUST reflect the accumulated conversation context**

The **frontend half is implemented**. The **server half does not exist**, and no id is ever minted, so the first half can never do anything either.

### What the code actually does

`lib/Controller/NaviController.php:94-101`:

```php
$conversationId = (string) $this->request->getParam('conversationId', '');

$payload = $this->naviService->processQuery(query: $query, userId: $user->getUID());
$payload['conversationId'] = null;
if ($conversationId !== '') {
    $payload['conversationId'] = $conversationId;
}
```

- The controller **never mints** a `conversationId`. It hardcodes `null` and then echoes back only a value the client already sent.
- `NaviService` never mints one either — `grep -n 'uuid\|Uuid\|uniqid\|random_bytes\|bin2hex' lib/Service/NaviService.php` returns **nothing**.
- `processQuery()` is not even **passed** the `conversationId` (`query:` and `userId:` only), so there is no channel by which context could accumulate. `NaviService.php:63` describes the value as something that "round-trips through `processQuery`" — it does not round-trip through it; the controller reattaches it afterwards.

`src/views/dashboard/widgets/NaviAnalyticsWidget.vue` holds up its end: `conversationId: null` initially (`:115`), sent on every request (`:141`), and adopted from the response when truthy (`:143-144`).

### The consequence

Turn 1 sends `null` → the server returns `null` → the widget adopts nothing → turn 2 sends `null` → forever. **The id is null for the entire life of the conversation on a live instance**, and the second bullet of the scenario ("the response MUST reflect the accumulated conversation context") has no implementation at all. Every Navi answer is computed from the current question alone.

### Why no test catches it, and why the green one is not reassurance

`tests/e2e/spec-coverage/dashboard.spec.ts:313` ("a follow-up question carries the conversationId from the first answer") asserts this scenario, and it passes — but it passes over a `page.route('**/api/navi/query')` interception that **supplies the `conversationId` the server does not**. That was a deliberate, documented choice (see the block comment above the test): without the stub the widget's half of the contract is unobservable, because a widget that faithfully propagates the id and one that drops it produce byte-identical traffic when the id is always `null`.

So the test is honest about what it covers — the **frontend** requirement — and is worth keeping as-is. But it should not be read as covering the scenario, and this issue is the missing half.

### Suggested fix

1. Mint an id in `NaviController::query()` when the client sends none (e.g. `bin2hex(random_bytes(16))`), so the widget has something to adopt on turn 1.
2. Pass `conversationId` **into** `NaviService::processQuery()` and give it somewhere to keep per-conversation state, so the second bullet can be satisfied.
3. Then replace the stub in `dashboard.spec.ts:313` with a live round trip, which becomes possible for the first time, and drop the block comment recording this mismatch.

If accumulated context is **not** wanted, the honest alternative is to amend the scenario — but it should be one or the other, not a spec bullet with no implementation and a green test over a stub.

Found while triaging the E2E failure on #769; the failure itself was unrelated (fixture cost, not this).

rubenvdlinde:
Closes #778.

## The defect

`openspec/specs/dashboard/spec.md#conversational-follow-up` requires two things. The frontend half was implemented. The server half did not exist:

- `NaviController::query()` hardcoded `$payload['conversationId'] = null` and then echoed back only a value the client had sent. **Nothing ever minted one.**
- `processQuery()` was never even *passed* a `conversationId`, so there was no channel by which context could accumulate.

Turn 1 sent `null` → the server returned `null` → the widget adopted nothing → turn 2 sent `null` → forever. The id was **null for the entire life of every conversation on a live instance**, and *"the response MUST reflect the accumulated conversation context"* had no implementation at all.

## What this does

**Mints the id.** `resolveConversationId()` accepts a client value only when it matches `/^[0-9a-f]{32}$/`; anything else — absent, empty, wrong shape, not a string — is discarded and a fresh `bin2hex(random_bytes(16))` is minted. ⚠️ The validation is not cosmetic: an unvalidated client string becoming a cache key is a key-injection primitive.

**Gives the conversation somewhere to live.** New `NaviConversationStore` over `ICacheFactory::createDistributed()`, TTL 3600 s, max 10 retained turns. **No new schema** — that is the design intent recorded in `NaviService`'s file docblock and `openspec/changes/dashboard/design.md §1`.

**Makes the answer actually use it.** `processQuery(query, userId, conversationId)` reads the history, asks the store what carries forward, uses it, then appends the turn.

## 🔒 Cross-user isolation — two independent barriers, each proven load-bearing

The endpoint is `#[NoAdminRequired]`, so a conversation id that is merely guessable would otherwise be a read primitive on someone else's conversation.

1. The cache key is `hash('sha256', $userId."\0".$conversationId)` — a conversation id alone addresses nothing.
2. The owning user id is stored **in** the record and re-checked on read; a mismatch logs and returns empty.

Neither is decorative, and that was verified rather than assumed:

| barrier disabled | what fails |
|---|---|
| key not user-scoped | the store holds 1 entry where 2 are required for one id used by two users |
| owner check removed | the leaky-cache test fails |
| **both** | Bob literally reads Alice's answer — `Found 4 records for "And what about last month?"` |

- `testConversationIsNotReadableByAnotherUser` — Alice and Bob over the **same** cache and the **real** store; Bob replays Alice's exact id and must get the clarification.
- `testConversationRecordOwnedByAnotherUserIsIgnored` — a deliberately leaky cache returns Alice's record for *every* key (the worst case a collision or a poisoned store produces); Bob still gets the clarification.

## The test that proves the second spec bullet

`NaviServiceTest::testFollowUpWithoutIntentKeywordInheritsPreviousIntent`. Turn 1 counts requests (4 fixture rows); turn 2 is *"And what about last month?"*, which matches **no** intent keyword. The answer must not contain the clarification and must contain `Found 4 records`.

Its control, `testQueryWithoutIntentKeywordGetsClarification`, asks the same wording **outside** any conversation and gets the clarification — so the inherited answer provably came from the store, not from the wording.

Second observable: `testFollowUpInheritsPreviousTurnSubject` — *"And how many of those are overdue?"* after a requests question must answer `Found 4 records`, not `Found 1 records` (the lead fallback). Control: `testFollowUpWithoutConversationFallsBackToLeads`.

**Positive controls run, not assumed.** With `history` forced to `[]` at the call site, both context tests fail (2 failures). Restored, they pass. **This is not a store-and-never-read no-op**, which would have satisfied bullet 1 and left bullet 2 exactly as unimplemented as it was.

## ⚠️ A correction to the issue's own premise

#778 says *"And how many of those are overdue?"* currently detects as intent `unknown`. **It does not** — `/\b(…|how many)\b/` matches, so it detects as `count`, and the visible defect is the **subject** (it counts leads). Both inheritance paths are therefore covered separately rather than claiming the issue's exact sentence. **No spec wording was touched** — `#conversational-follow-up` was implemented as written.

## Deliberate scope addition, flagged

`TicketService` gains `detectTypeInText()` and a `TYPE_VOCABULARY` constant; the NL/EN subtype words moved out of `NaviService` to sit with the `TYPE_*` constants they describe. **It also gained the plural `requests`**, which the old `\brequest\b` pattern never matched — so `"How many requests are open?"` used to silently count **leads**. That changes behaviour beyond the issue and is covered by `TicketServiceTest::testDetectTypeInTextRecognisesSubtypeVocabulary`.

The extraction was also what kept PHPMD honest: the additions pushed `NaviService` WMC 48 → 71 against a threshold of 50, and `phpmd.baseline.xml` explicitly forbids regenerating itself. Fixed by extraction, as that file prescribes — `NaviService` is back at **48**, its exact baseline. **No baseline entry, no threshold bump.**

## E2E

`tests/e2e/spec-coverage/dashboard.spec.ts:313` passed over a `page.route()` interception that **supplied the `conversationId` the server did not**. That stub is replaced by a live round trip, the block comment recording the mismatch is removed, and the header note is corrected (3 → 2 stubbed Navi scenarios).

⚠️ **Zero document loads added** — the same single navigation, plus two XHR round trips against a deterministic (non-LLM) service. On this suite a `page.goto()` costs 21–26 s because N workers queue on one container's static-asset service, and a recent fixture spent 58.1 s of a 60 s budget that way. Timeouts are 30 s bounds matching the sibling live Navi test, not delays. **No widened timeout, no `networkidle`, no retries, no `mode: 'serial'`, no skip.**

The other two Navi e2e stubs stay: the CI seed makes those branches unreachable, which is unrelated to this issue.

## Verification

| check | result |
|---|---|
| PHPUnit `Unit Tests` | **1646 tests, 5175 assertions, 0 failures** |
| phpcs | clean on all 4 changed `lib/` files |
| phpmd | clean — `NaviService` back to its exact baseline WMC |
| phpstan | `[OK] No errors` |
| psalm | `No errors found!` |
| `@spec` anchors | all resolve (canonical `check_spec_anchors.py`, exit 0) |

⚠️ Not verified: the `.ts` was not compile-checked — this worktree has no `node_modules`. It was reviewed by hand and uses only `waitForRequest`/`waitForResponse`/`postDataJSON`, all already used elsewhere in the same file. CI will be the first compile.

### DIFF
--- a/lib/Controller/NaviController.php
+++ b/lib/Controller/NaviController.php
@@ -8,6 +8,12 @@
  * `{ query, conversationId? }`, delegates to the service, and returns a
  * structured response envelope (`resultType` + optional chart/table data).
  *
+ * The controller owns the conversation identifier: it validates the
+ * client-supplied value against the minted shape and mints a fresh one
+ * whenever the client sends nothing or sends something else, so the response
+ * always carries a usable identifier and an attacker-chosen string never
+ * reaches the conversation store as a key.
+ *
  * @category Controller
  * @package  OCA\Pipelinq\Controller
  *
@@ -19,7 +25,7 @@
  *
  * @link https://pipelinq.nl
  *
- * @spec openspec/changes/dashboard/tasks.md#task-1.1
+ * @spec openspec/specs/dashboard/spec.md#requirement-navi-ai-analytics-widget-req-dash-001
  *
  * SPDX-FileCopyrightText: 2026 Conduction B.V. <info@conduction.nl>
  * SPDX-License-Identifier: EUPL-1.2
@@ -46,10 +52,20 @@ use Psr\Log\LoggerInterface;
  * logic lives in `NaviService`; this controller stays under 10 lines per
  * method per ADR-003.
  *
- * @spec openspec/changes/dashboard/tasks.md#task-1.1
+ * @spec openspec/specs/dashboard/spec.md#requirement-navi-ai-analytics-widget-req-dash-001
  */
 class NaviController extends Controller
 {
+    /**
+     * The only accepted shape for a conversation identifier: the 32 lower-case
+     * hex characters produced by `bin2hex(random_bytes(16))`. A value that does
+     * not match is treated as absent, so a caller can never choose the string
+     * that becomes part of a conversation-store key.
+     *
+     * @var string
+     */
+    public const CONVERSATION_ID_PATTERN = '/^[0-9a-f]{32}$/';
+
     /**
      * Constructor.
      *
@@ -74,9 +90,14 @@ class NaviController extends Controller
      * Missing/empty `query` -> 400. Unauthenticated -> 401. Underlying
      * failure -> 500 with a static message (no `getMessage()` leak).
      *
+     * The returned payload always carries a non-null `conversationId`: the
+     * client's value when it matches the minted shape, a freshly minted one
+     * otherwise. That identifier is handed to the service so follow-up turns
+     * can be answered against the accumulated conversation.
+     *
      * @return JSONResponse Response envelope or error.
      *
-     * @spec openspec/changes/dashboard/tasks.md#task-1.1
+     * @spec openspec/specs/dashboard/spec.md#conversational-follow-up
      */
     #[NoAdminRequired]
     public function query(): JSONResponse
@@ -91,14 +112,14 @@ class NaviController extends Controller
             return new JSONResponse(['message' => 'Missing query'], Http::STATUS_BAD_REQUEST);
         }
 
-        $conversationId = (string) $this->request->getParam('conversationId', '');
-
         try {
-            $payload = $this->naviService->processQuery(query: $query, userId: $user->getUID());
-            $payload['conversationId'] = null;
-            if ($conversationId !== '') {
-                $payload['conversationId'] = $conversationId;
-            }
+            $conversationId = $this->resolveConversationId(raw: $this->request->getParam('conversationId', ''));
+            $payload        = $this->naviService->processQuery(
+                query: $query,
+                userId: $user->getUID(),
+                conversationId: $conversationId
+            );
+            $payload['conversationId'] = $conversationId;
 
             return new JSONResponse($payload);
         } catch (\Throwable $e) {
@@ -109,4 +130,29 @@ class NaviController extends Controller
             return new JSONResponse(['message' => 'Navi unavailable'], Http::STATUS_INTERNAL_SERVER_ERROR);
         }//end try
     }//end query()
+
+    /**
+     * Accept a client-supplied conversation identifier, or mint a fresh one.
+     *
+     * Only the shape this controller itself mints is accepted. Anything else —
+     * absent, empty, wrong length, non-hex, or not a string at all — yields a
+     * new identifier, so an unvalidated caller string is never used as part of
+     * a cache key.
+     *
+     * @param mixed $raw The raw request parameter.
+     *
+     * @return string A 32-character lower-case hex identifier.
+     *
+     * @throws \Random\RandomException When the CSPRNG is unavailable.
+     *
+     * @spec openspec/specs/dashboard/spec.md#conversational-follow-up
+     */
+    private function resolveConversationId(mixed $raw): string
+    {
+        if (is_string($raw) === true && preg_match(self::CONVERSATION_ID_PATTERN, $raw) === 1) {
+            return $raw;
+        }
+
+        return bin2hex(random_bytes(16));
+    }//end resolveConversationId()
 }//end class
--- a/lib/Service/NaviConversationStore.php
+++ b/lib/Service/NaviConversationStore.php
@@ -1,0 +1,255 @@
+<?php
+
+/**
+ * Pipelinq NaviConversationStore.
+ *
+ * Ephemeral per-user storage for Navi conversations. A conversation is a short
+ * list of turns (`query` + detected `intent`) held in the Nextcloud distributed
+ * cache under a TTL, which is what lets a follow-up question be answered in the
+ * context of the question before it.
+ *
+ * Two independent barriers keep one user's conversation out of another's
+ * answers, because the Navi endpoint is reachable by any authenticated user and
+ * the identifier travels in the request body:
+ *   1. the cache key is derived from the user id AND the conversation id, so
+ *      replaying someone else's identifier addresses a different record;
+ *   2. the owning user id is stored inside the record and re-checked on read,
+ *      so a record that somehow surfaces under the wrong key is still refused.
+ *
+ * No register or schema is involved: conversation context is ephemeral by
+ * nature and losing it on a cache flush costs nothing but the thread of one
+ * chat. Growth is bounded by the TTL and by a maximum number of retained turns.
+ *
+ * @category Service
+ * @package  OCA\Pipelinq\Service
+ *
+ * @author    Conduction <info@conduction.nl>
+ * @copyright 2026 Conduction B.V.
+ * @license   EUPL-1.2 https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12
+ *
+ * @version GIT: <git_id>
+ *
+ * @link https://pipelinq.nl
+ *
+ * @spec openspec/specs/dashboard/spec.md#conversational-follow-up
+ *
+ * SPDX-FileCopyrightText: 2026 Conduction B.V. <info@conduction.nl>
+ * SPDX-License-Identifier: EUPL-1.2
+ */
+
+declare(strict_types=1);
+
+namespace OCA\Pipelinq\Service;
+
+use OCP\ICache;
+use OCP\ICacheFactory;
+use Psr\Log\LoggerInterface;
+
+/**
+ * User-scoped, TTL-bounded store for Navi conversation turns.
+ *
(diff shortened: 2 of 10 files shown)


rubenvdlinde:
## 🔴 CI found a real defect in this PR, and the **live e2e test this PR introduced** is what caught it

`dashboard.spec.ts:333` failed twice (initial + retry #1):

```
Error: the follow-up must be answered from the accumulated context
Expected substring: not "I am not sure how to answer that yet"
Received string:        "I am not sure how to answer that yet. Try asking about leads, requests, or trends."
```

On a real instance the follow-up gets the **clarification** — i.e. the accumulated context is not there, which is precisely the spec bullet this PR claims to implement.

### Cause

`NaviConversationStore::cache()` calls `ICacheFactory::createDistributed()`. In `\OC\Memcache\Factory`:

```php
public function createDistributed(string $prefix = ''): ICache {
    $cache = new $this->distributedCacheClass(...);   // NULL_CACHE when none is configured
}
public function isAvailable(): bool {
    return $this->distributedCacheClass !== self::NULL_CACHE;
}
```

With no memcache configured, `createDistributed()` returns a **`NullCache`**: `set()` silently succeeds, `get()` returns null. The store is a black hole.

Confirmed two independent ways: the E2E log contains no memcache/redis at all, and the shared `ConductionNL/.github` workflows configure none. ⚠️ **This is not a CI quirk — a default Nextcloud install has no distributed cache**, so this PR as it stands would ship a feature that silently does nothing on most instances.

### 🔑 The lesson, which is the valuable part

**1646 green unit tests did not see this.** They injected a working cache double, so the property that actually mattered — *written in request A, readable in request B* — was never exercised. The one test that could see it is the live round trip this PR added to replace the old stub.

That is a direct instance of *green unit tests around a fix are not evidence the fix reached the behaviour*, and it is also the vindication of removing the stub: **the previous stubbed version of this very test would have passed over this bug**, because the stub supplied the `conversationId` the server did not.

### Being fixed, not worked around

Sent back for a **durable** store — the property to prove is cross-request survival on an instance with no memcache, asserted directly in a unit test with two store instances over one backing store. Likely shape: one record per user in user preferences (durable, no migration, no new schema, and per-user **by construction**, which makes the cross-user isolation structural rather than a check — the explicit owner check stays as defence in depth).

⛔ **Not** by re-stubbing the route, adding a retry, widening a timeout, or weakening the assertion. This test is currently the only thing in the suite that can see this class of bug.

**Do not merge until that lands.**

rubenvdlinde:
## Reviewed adversarially, pushed as `f06b1a816`. One correction to the control that was reported with it.

### What I re-verified rather than accepted

| claim | verdict |
|---|---|
| the e2e test is byte-identical | ✅ `git diff de91466cf f06b1a816 -- tests/e2e/` is **empty** |
| the durability proof models two requests | ✅ `buildStore()` mints a **fresh** store, `IConfig` mock, `ITimeFactory` and logger every call; only `$this->rows` — the test-case-level array standing in for the database — crosses. No in-process state can carry a value between the two halves, which is exactly the property that failed before |
| `append()` actually stores the new turn | ✅ `$history[] = $turn;` precedes the `MAX_TURNS` slice |
| a *different* conversation id cannot read the previous conversation | ✅ `isUsable()` compares `$record['conversationId'] !== $conversationId` — no context bleed, and there is a test for it |
| owner check retained | ✅ plus both isolation tests |
| 1655 / 5193 / 0 failures; phpcs, phpmd, phpstan, psalm, `@spec` anchors | ✅ all reproduced locally |

### Two hazards the mocked `IConfig` cannot see, checked against the server source

- **Column width.** `oc_preferences.configvalue` is declared `text`, not `varchar` (`core/Migrations/Version13000Date20170718121200.php:307-309`). A bounded record — 10 turns × 240 chars ≈ 2.7 KB — fits with room to spare.
- **Length caps in the NC 31+ path.** `IConfig::setUserValue()` now delegates to `IUserConfig`. `UserConfig`'s `*_MAX_LENGTH` constants bound **userId (64)**, **app (32)** and **key (64)** — **not the value**. `pipelinq` is 8 and `navi_conversation` is 17, both comfortable.

### 🔴 Correction — the positive control's second arm does not reproduce

Reported: *"mutating `read()` to behave like `NullCache` fails 7 tests; the old suite had **0 failures** under that same condition."*

The first half is exactly right — I reproduced **7**, and they are the right seven. **The second half is wrong.** Applying the equivalent mutation to `de91466cf` (force the cache read to return null, drop the new store test file) gives:

```
Tests: 1646, Assertions: 5173, Failures: 3
  NaviServiceTest::testFollowUpInheritsPreviousTurnSubject
  NaviServiceTest::testFollowUpWithoutIntentKeywordInheritsPreviousIntent
  NaviServiceTest::testConversationRecordIsBoundedByTurnsAndTtl
```

**3, not 0.**

The distinction matters, and it sharpens what was actually untested. That mutation models *"the read is dead"*, which is **stronger** than the real defect and which the old suite could partly see. The real defect was *"the write does not survive to the next request"* — and the old suite was blind to **that** specifically, because its cache double persisted in-process and **both turns ran on one `NaviService` instance**. A write followed by a read inside one test passed exactly as it should.

So the honest statement is not "the old suite caught nothing", it is: **the old suite could detect a dead read but had no way to detect a non-durable store, because nothing in it ever crossed an instance boundary.** That is precisely the gap the new `buildStore()`-per-request tests close, and it is why the fix to run each turn on its own service instance is as important as the storage change.

🔑 A control that fails N tests tells you the mutation was detectable. It does **not** tell you the *original* defect was — for that the mutation has to be the defect, not a stronger relative of it.

Un-marked WIP. Merging if CI agrees; the live e2e round trip is what will establish the real `IConfig` against the real database, which no unit test here claims to cover.

rubenvdlinde:
ℹ️ The `quality / Security (composer)` failure here is **not from this PR**. `CVE-2026-65954` (arbitrary code execution in `phpcsstandards/phpcsutils <1.2.3`, a transitive dev dependency) was published to the advisory database between #796's run at 19:13 (passed) and this one at 19:28 (failed), with no dependency change in either.

It affects `development` itself and blocks every open PR. Fix is up as #802 — five-line lock diff, A/B-verified as sniff-behaviour-neutral. Once that lands, rebase this branch and the cell clears.

Issue #805, opened 2026-08-11
Customer 360 has never worked — a fail-closed dead guard whose one-line fix turns it into an IDOR


rubenvdlinde:
## Summary

`GET /api/customer-360/summary` has **never returned data to anybody**, and the one-line fix for that turns it into an IDOR. Both halves are measured below.

## 1. The endpoint is dead

`Customer360Controller::canReadClient()` (`:142`) calls `ObjectService::find()` **positionally**:

```php
$object = $objectService->find($clientId, $register, $schema);
```

The upstream signature is `find($id, ?array $_extend, bool $files, $register, $schema, …)`, so `$register` lands in `?array $_extend`:

```
Customer360Controller: client read-guard failed |
OCA\OpenRegister\Service\ObjectService::find(): Argument #2 ($_extend) must be of type
?array, string given, called in .../Customer360Controller.php on line 142
```

`catch (Throwable) { … return false; }` swallows it, so the guard denies unconditionally.

**Measured live** (single-owner rig, NC 34.0.0, two accounts, Basic auth, status codes printed). A `client` object created via OpenRegister by `pqvictim` → **201**:

| caller | HTTP |
|---|---|
| `pqvictim` — **created the client** | **404** `Client not found` |
| `pqattacker` | **404** `Client not found` |

Same root cause as #794 and as the `ContractController` half of #801 (fixed in #804), but the **opposite failure direction**: that one fell into a 404 branch, this one falls into a fail-closed catch.

## 2. 🔴 Why the obvious fix is dangerous

Making the call named — exactly what #804 does for `ContractController` — makes the guard evaluate its predicate for the first time. The predicate is:

```php
return $object !== null;
```

**That is an existence check, not an authorisation check.** It is true for every client that exists, for every authenticated caller. So repairing the call would convert an endpoint that denies *everyone* into one that grants *everyone* a full customer-360 read — contactmomenten, contracts, invoices, consent state — on **any** `clientId`.

The endpoint is `#[NoAdminRequired]`, `client` declares **no owner field**, and pipelinq declares an `authorization` block on 1 of ~164 schemas, so OpenRegister's `PermissionHandler` grants unconditionally underneath (ConductionNL/.github#372, openregister#2011). **The controller is the only layer, and its guard is an existence check.**

⚠️ **The fatal call is currently the only thing suppressing the wrong predicate.** A crash is not neutral here — it is load-bearing.

Compounding: the method's docblock is the most emphatic security comment in the file —

[...]

Issue #801, opened 2026-08-11
SECURITY: 23 #[NoAdminRequired] endpoints are unguarded IDORs — incl. cross-app invoice mint and address-book exfiltration (gate-7 reports 0; .github#365)


rubenvdlinde:
## Summary

**23 `#[NoAdminRequired]` endpoints accept an object id off the request and act
on it with no per-object authorisation** — including one that mints a real
invoice in shillinq against a contract the caller never owned, one that stamps an
attacker-controlled URL onto any contactmoment, and one that exfiltrates any
client record into the caller's own Nextcloud address book.

gate-7 reports **`0`** for this app and always has.

## Why this was never reported

**`.github#365`** — gate-7 accepts `Http::STATUS_UNAUTHORIZED` / `401` as an
authorisation guard. **49 of pipelinq's 82 controller files** open with the
house-style `if ($user === null) { … 401 }` preamble, which is *authentication*.
Everything behind it is exempt for free. Compounding: `.github#372` — pipelinq
declares an `authorization` block on **1 of ~164 schemas**, so OpenRegister's
`PermissionHandler` grants unconditionally
(ConductionNL/openregister#1955, #2011).

## Measurement

Canonical gate package `ConductionNL/.github` @ `112d4c9` (post-#368), over the
tracked `lib/Controller` tree at `origin/development` `67cb89b89`:

| checker | findings |
|---|---:|
| canonical (what CI reports today) | **0** |
| same run, `UNAUTHORIZED`/`401` dropped from `_GUARD_BODY_RE` | **49** |

Of those, 26 take a caller-supplied object id; hand-verification confirmed
**23 REAL**, 3 not-a-DOR.

⚠️ Four of these (the `LoyaltyController` set) are the ones `.github#365` already
names. **This issue supersedes that enumeration — the same file holds five more.**

## Positive control

pipelinq **does** implement per-object authorisation:
`lib/Controller/ContractController.php:157` —
`if ($this->isAuthorized(uid: $user->getUID(), contract: $contract) === false) { … STATUS_FORBIDDEN }`,
backed by `:275` `if (((string) ($contract['ownerId'] ?? '')) === $uid) { return true; }`.
So every absence below is a real omission, not a missing convention.

## 🔴 Worst — `POST /api/handoff/contract/{id}/send-to-invoicing`

`SemanticHandoffController::sendContractToInvoicing` (`:256`)

It is a write, it is irreversible from this app's side, and **it crosses an app
[...]

PR #808, opened 2026-08-11
fix(listeners): revive three dead listeners and four lost object ids — method_exists() is false for every OR magic accessor (#807)

### LINKED ISSUE #807: Three listeners are permanently dead — POS loyalty points, expense approval and Berichtenbox status all probe magic accessors with method_exists()
An agent is live in this repo, so this is filed rather than fixed. All line numbers verified against `origin/development` with `git show`, not a local working tree.

## The mechanism

`method_exists()` is **false** for any accessor Nextcloud's `Entity` serves through `__call()`. OpenRegister's `ObjectEntity` declares `getSchema()`/`getUuid()` as `@method` docblocks only.

Measured live against the server's own `Entity.php`:

| subject | `method_exists` | `is_callable` |
|---|---|---|
| `ObjectEntity::getSchema` / `getUuid` | **false** | true |
| `ObjectEntity::getObject` — concrete control | true | true |

This repo **already knows this** — `lib/Service/PosTransactionService.php:1452` carries a comment stating exactly it. The knowledge did not reach these call sites.

## Tier 1 — three listeners are entirely dead

Each guard sits **before** the app-config fallback that would otherwise rescue it, so the fallback never runs.

| site | effect |
|---|---|
| `lib/Listener/PosTransactionCompletedListener.php:129` | `isPosTransaction()` always false → the **entire POS loyalty-points award path** never runs |
| `lib/Listener/ExpenseApprovalListener.php:222` | `isExpense()` always false → no `ExpenseApprovedEvent`, no AP webhook, no `apSyncStatus` write |
| `lib/Listener/BerichtenboxZaakStatusListener.php:170` | `isZaakUpdate()` always false → no citizen-facing Berichtenbox notification on a zaak update. `ObjectUpdatedEvent::getNewObject()` is hard-typed `: ObjectEntity`, so this is certain |

## Tier 2 — returned ids silently lost

The row **is** persisted; only the returned handle is null.

`SlaEngineService.php:1344` (SLA-breach audit id) · `CtiDispositionService.php:225` (returns null) · `ContactmomentService.php:201` (returns null) · `BsnAuditService.php:151` (AVG/BSN audit handle, 4 callers)

## Not defects — please do not "fix" these

`SlaDeadlineSweepJob:340`, `SlaPolicyController:252`, `InitSlaStatus:276`, `SlaEngineService:1451`, `ActivityTimelineService:498,923`, `EntityActivityService:537` all have a **working fallback**: `ObjectEntity::getObject()` always injects the uuid under `id` (`array_merge(['id' => $this->uuid], $objectData)`), so `$data['id']` already holds. Dead code, no runtime effect.

## Why the suite is green, and what a real fix needs

`tests/Stubs/Db/ObjectEntity.php` declares `getUuid` (:50) and `getSchema` (:62) **concretely**, and its docblock justifies this for `onlyMethods()`. That inverts the exact predicate under test: the probe is true in the suite and false in production. **A fix without reshaping the double proves nothing.**

Write the prediction down first — how many tests go red when the fix is reverted — then check it.

`decidesk/tests/Stubs/Db/ObjectEntity.php` is the reference: it extends `Entity`, declares only genuinely concrete methods, and carries a "SIGNATURE PARITY CONTRACT" header naming this failure mode. decidesk has one residual defect; this repo has five.

## Two traps when fixing

1. **`is_callable()` is not a membership test.** Measured: it is **true for any name** on a `__call` class, and the call then raises `BadFunctionCallException`. Swapping the probe produces an always-true guard, not a working one — make the call exception-safe in the same edit.
2. **`getId()` is magic too.** `method_exists($x, 'getId')` is false on every OR entity. Do not fall back to it. This repo has three `method_exists($x, $variable)` sites as well (`CtiService:1021`, `SemanticHandoffService:216`, `BlastService:1354`) which a literal-name grep cannot see.

Fleet context and the full 18-repo table: `fleet-board/findings/method-exists-sweep.md`.

rubenvdlinde:
Closes #807.

## The measurement

Run on a live Nextcloud 34 against the **real** `OCA\OpenRegister\Db\ObjectEntity`, with concrete controls in the same table so the probe itself is shown to work:

| accessor | `method_exists` | `is_callable` | `property_exists` | direct call |
|---|---|---|---|---|
| `getObject()` | true | true | true | ok *(concrete control)* |
| `jsonSerialize()` | true | true | true | ok *(concrete control)* |
| `getSchema()` | **false** | true | true | ok |
| `getUuid()` | **false** | true | true | ok |
| `getId()` | **false** | true | true | ok |
| `getNoSuchThingAtAll()` | false | **true** | **false** | `BadFunctionCallException` |

The last row is why **`is_callable()` is not the fix**: it is true for a name the entity has never heard of, so a probe swap makes the branch unconditionally true and the call then throws. `Entity::getter()` decides with `property_exists()`; where a **value** is wanted the correct shape is a direct call in `try/catch` plus `is_scalar` — copied from `shillinq/lib/Service/ListenerSchemaResolver.php:163`.

## What is repaired

`lib/Util/EntityAccessorTrait::readEntityValue()`, used at 8 sites:

| site | effect while broken |
|---|---|
| `ExpenseApprovalListener::isExpense` | no `ExpenseApprovedEvent`, no AP webhook, no `apSyncStatus` write — ever |
| `PosTransactionCompletedListener::isPosTransaction` | the entire POS loyalty award path |
| `PosTransactionCompletedListener::getEntityUuid` | `posTransactionId` empty whenever the payload had no reference |
| `BerichtenboxZaakStatusListener::isZaakUpdate` | see the caveat below |
| `SlaEngineService:1344` · `CtiDispositionService:225` · `ContactmomentService:202` · `BsnAuditService:151` | `saveObject()` is typed `: ObjectEntity`, so each wrote its row and returned `null` / `''` instead of the uuid |

## Two corrections to the issue

**1. Fixing the probe does NOT revive the POS loyalty path on its own.** The listener reads `klantId` / `customerId` / `contactUid`, `totaalbedrag`, `voltooidOp`, `transactieId`, `kanaal`. The `posTransaction` schema declares **none** of them, and `grep` finds `totaalbedrag` / `voltooidOp` / `transactieId` **nowhere else in the app** — nothing writes them. The declared fields are `customer` (documented as *"UUID reference to a pipelinq contact"*, which is exactly what `processPosTransaction()` calls its `$klantId`), `total`, `settledAt`, `reference`, `terminalId`. With only the probe fixed the listener moves from *never fires* to *fires, finds no customer, returns* — and had it got past that, it would have awarded points on `amount = 0`. The schema names are now read first and the legacy names kept as a fallback.

**2. `BerichtenboxZaakStatusListener` still cannot fire, and that is deliberate.** `zaak_schema` appears exactly once in the whole repository — on the line that reads it. It is absent from `SchemaMapService::SCHEMA_KEYS`, absent from `SettingsService`, and the merged register (monolith + 30 `register.d` fragments, 96 schemas) declares no `zaak` schema at all. So `$zaakSchema` is `''` and the comparison fails regardless. Deciding which schema is a "zaak" is a product/configuration call, not something to invent inside a defect fix; the second kill switch is documented in the code and reported on #807.

`PosTransactionCompletedListener:178` (`getEntityUuid`) was briefed to me as a confirmed false positive because `getObject()` injects `id => uuid`. **It is not one** — the caller reads `transactieId` / `transactionId`, never `id`, so the injection does not reach it. Fixed. The other seven "not a defect" sites in #807 were each re-read and confirmed harmless.

## Why 2 200 green tests could not see any of this

`tests/Stubs/Db/ObjectEntity.php` declared `getUuid()` and `getSchema()` **concretely**, so `method_exists()` was **true** in the suite and **false** in production — the double inverted the exact predicate under test, and its docblock justified the inversion as a feature. It now `extends Entity` and declares only what production declares (`getObject`, `jsonSerialize`), with a signature-parity contract in its header. The five test files that mocked those accessors build real entities through the magic setters. `tests/Stubs/Event/ObjectUpdatedEvent.php` gains the `getObject()` production has (`openregister lib/Event/ObjectUpdatedEvent.php:71`) and it lacked.

`EntityAccessorTraitTest::testTheDoubleReproducesProductionsMagicAccessorShape` is the standing guard: it asserts `method_exists` **true** for the two concrete accessors and **false** for the two magic ones, so the double cannot drift back without a red cell.

## Revert predictions — written before running, then measured

| revert | predicted | observed |
|---|---|---|
| `isExpense` → `method_exists` | **4** (2 unit + 2 integration, named) | **4**, exactly those names |
| `isPosTransaction` → `method_exists` | **2** | **2** |
| `getEntityUuid` → `method_exists` | **1** | **1** |
| POS vocabulary → legacy keys only | **2** | **2** |
| `ContactmomentService` → `method_exists` | **1** | **1** |
| faithful stub → old concrete stub | **29** | **28** |

Every revert used `Edit`, never a script, and the reverted line was verified by grep before each run.

⚠️ **The last row is off by one and the cause is worth recording rather than rounding away.** I predicted `EntityAccessorTraitTest` would lose 4 of its 5 tests; it lost 3. `testAnUnknownAccessorYieldsAnEmptyStringInsteadOfThrowing` **passes under both stubs**: the old stub is a plain class with no `__call`, so an unknown accessor raises `Error: Call to undefined method`, which `readEntityValue`'s `catch (Throwable)` swallows just as it swallows `BadFunctionCallException`. That test is therefore insensitive to which stub is installed — a weaker guard than I assumed. The parity control above is the one that actually pins the double.

Also note what stays **green** under the first revert: the four tests asserting the listener does *nothing* (`testIgnoresNonExpenseSchema`, `testIgnoresNonApprovedExpense`, `testNoopWhenUnconfigured`, `testIdempotentSkipsAlreadySynced`). They pass because the feature is dead — which is precisely the shape that kept this invisible.

## Checks

`lint`, `phpcs`, `phpmd`, `psalm`, `phpstan` all exit 0. **2 211 tests / 7 549 assertions / 0 failures** (2 200 → 2 211: +11 new).

Two pre-existing `phpcs` warnings fixed on the way (missing class-level `@spec` on `PosTransactionCompletedListener` and on `BerichtenboxZaakStatusListener::handle`).


rubenvdlinde:
## Verified — and one blocker that the revival itself creates

I re-measured this independently rather than trusting the write-up. The diagnosis, the instrument choice and the comparand analysis all hold. One thing does not, and it is the reason I am not merging.

### What I confirmed first-hand

**The probe defect is real, at the class level.** `OCA\OpenRegister\Db\ObjectEntity` on `origin/development` declares `getObject()` and `jsonSerialize()` concretely and declares the properties `uuid`, `register`, `schema`, `object` — but no `getSchema()`, `getUuid()` or `getId()` method. So `method_exists` is false and `property_exists` is true for exactly the accessors this PR repairs, and the concrete controls in the table are genuine controls.

**The comparand is a numeric id, not a slug — so the probe fix is sufficient here.** Measured on the live instance, not inferred:

| | value |
|---|---|
| `pipelinq.posTransaction_schema` | `434` |
| `pipelinq.expense_schema` | `507` |
| `_schema` on all 161 rows of the posTransaction shard | `434` |
| `_schema` on all 5 rows of the expense shard | `507` |
| `oc_openregister_schemas` slug for 434 / 507 | `posTransaction` / `expense` |

The chain that guarantees it is `SettingsMapBuilder::addSchemaToMap()` (`$schemaArray['id'] ?? $schemaArray['uuid']`, keyed by slug) into `SettingsLoadService::applySchemaConfig()` (`<slug>_schema = <that value>`) into `SchemaMapService::buildSchemaMap()`, which keys its map on the same values. Both arms of all three guards therefore compare id-to-id. Adding `SchemaMapper::find($raw)->getSlug()` would **break** this app, and the docblock says so. This is the mirror of the decidesk case and the distinction is drawn correctly.

**The POS vocabulary correction is right.** The live `posTransaction` schema declares `customer`, `total`, `settledAt`, `reference`, `terminalId` and does not declare `klantId`, `totaalbedrag`, `voltooidOp`, `transactieId` or `kanaal`. Its `status` enum is `draft|parked|confirmed|settled|refunded`, so `settled` is the live completion value and the two legacy values are correctly kept only as extras.

**`getEntityUuid` really was not a false positive.** The `getObject()` injection of `id => uuid` cannot reach it, because the caller reads `reference` / `transactieId` / `transactionId` and never `id`. The correction to the brief is correct.

**The fix is load-bearing.** My own revert prediction, written before running: reverting `isPosTransaction` to `method_exists` fails 3 tests — the two named in the body plus the numeric-id comparand test added later. Observed exactly 3, and exactly the three that assert the listener *acts*. Reverted with `Edit`, verified by grep, restored to a clean tree.

**The double is fixed in the same commit**, which is what stops a correct fix reading as a regression, and the parity test asserts a backing property for all four magic accessors, not just the method names.

### The blocker

`ExpenseApprovalListener` was dead. Reviving it arms a re-entrant loop that nothing in this repo can currently see.

`dispatchApproval()` calls `persist()`, which calls `ObjectService::saveObject(...)` **without `silent: true`**. `saveObject`'s `$silent` is documented as *"whether to skip audit trail creation and events"* and defaults to `false`, and `MagicMapper::update()` suppresses the lifecycle event only when `SystemOperationContext::isActive()`, which is not the case here. So that write dispatches `ObjectUpdatedEvent` on the same expense object, which re-enters `handle()`:

- `isExpense` — still true
- `status` — still `approved`
- `apSyncStatus` — now `pending`, and the idempotency guard only short-circuits on `=== 'synced'`
- `shouldDispatch()` — still true

so `dispatchApproval()` runs again, persists `pending` again, and recurses. The `synced` write that would terminate the loop is never reached, because the recursion happens before it. The failure path is the same shape: `failed` is not `synced` either, and each level issues another outbound webhook call with retries.

This is bounded today only by `shouldDispatch()` being false when no `https://` webhook URL is configured — which is the default, and is why it is invisible. It arms the moment anyone configures the AP webhook. The unit suite cannot see it because the `saveObject` double does not dispatch events.

This is also what the red gate-61 cell is pointing at. The full-scope count on this branch is 8; 2 of them are this PR's and 6 are pre-existing (`DealCreatedListener`, `DealUpdatedListener`, `ProjectCreationListener`, `ProjectPhaseStatusListener`, `SlaObjectCreatedListener`, `SlaObjectUpdatedListener`). The 2 here are new in a way the other 6 are not: `persist()` already existed on `development`, but `isExpense()` was permanently false, so nothing ever reached it. The gate is correct and the finding is newly load-bearing.

### What I think closes it

1. **The re-entrancy**, which is the outage risk: the listener's own bookkeeping write must not re-broadcast. `silent: true` on the `persist()` save is the smallest correct change; tightening the idempotency guard to also short-circuit on `pending` is a reasonable belt-and-braces but does not on its own stop the `failed` path.
2. **gate-61**, which is the architectural half: the synchronous outbound webhook with retries inside an OpenRegister write is the category ADR-078 exists for. pipelinq has no `ListenerDeferralService` and 33 background jobs, so deferral is a real piece of work — but an inline annotation here would be buying the gate's silence rather than answering it, and none of the four closed categories honestly fits a remote HTTP call on the write path.

Everything else in this PR — the trait, the stub parity contract, the POS listener, the Berichtenbox listener's documented second kill switch, and the four `saveObject()` uuid reads — verified clean and I would land those today. It is specifically the expense path that needs the deferral question answered before it goes live.

I have deliberately not touched the `zaak_schema` gap: `zaak` (schema 744) exists on this instance but is not in pipelinq's register 16, and `zaak_schema` app-config is empty. Deciding which schema is a zaak is a product call and inventing it here would be authoring the requirement.

rubenvdlinde:
## The re-entrancy blocker is resolved. A second, smaller one is now measured — and it is why this should not be merged yet.

### 1. The loop is real, and the remedy this PR was held for would not have worked

`dispatchApproval()` → `persist()` → `saveObject()` → `ObjectUpdatedEvent` on the same expense →
back into `handle()`, where the idempotency guard short-circuits only on `=== 'synced'` and the
first re-entry sees the `pending` that `persist()` just wrote. Confirmed.

**But `silent: true` does not fix it**, and that was measured rather than assumed:

| what `$silent` gates | where |
|---|---|
| audit-trail row creation | `openregister lib/Service/Object/SaveObject.php:3468`, `:5445` |
| `updateInverseRelations()` | `:3489`, `:5515` |
| **the `ObjectUpdatedEvent` dispatch** | **not gated by `$silent` anywhere** |

The dispatch is `lib/Db/MagicMapper.php:8997` and is withheld by exactly one predicate —
`suppressLifecycleEvents()` → `SystemOperationContext::isActive()`, which a listener is not in.
⚠️ `ObjectService::saveObject()`'s own docblock (`:1173`) says *"skip audit trail creation **and
events**"*. The parameter is documented to do something it does not do.

`ObjectService::runAsSystem()` would suppress it, and its docblock even names *"event listeners
reacting to webcron-created objects"* — but its next paragraph says *"never wrap handling of
user-supplied request data"*, and an expense write-back is exactly that. Wrong trade for a status
field.

**Shipped instead:** an in-flight UUID set released in `finally`. Static rather than per-instance,
because Nextcloud resolves the listener from the container per dispatch, so a re-entrant dispatch
is not guaranteed to reach the same object.

**The control, with the number written down before it was run:** every existing ObjectService
double here captures `saveObject()` and stops, which is why the suite could be green over an
unbounded loop. The new double re-dispatches the way the mapper does, with a depth cap so an
unguarded listener fails with a countable number rather than a stack overflow.

| arm | predicted | observed |
|---|---|---|
| guard present | 1 `dispatchApEvent` | **1** |
| guard removed, re-dispatch capped at 3 | **4** | **4** |

Re-run after the extraction that fixed the phpmd complexity regression, because delegation is
where a test's teeth get lost — same prediction, same result, and `grep -c` confirmed which line
was actually out at each arm.

### 2. The new blocker: merging this reds gate-61 on `development`

Predicted locally against `origin/development` with the canonical checker, before merging:

```
FAIL  lib/Listener/ExpenseApprovalListener.php (ObjectCreatedEvent): performs a write (saveObject)
      inside another object's write, and it neither defers via ListenerDeferralService nor carries
      `@listener-placement inline <category> — <reason>` on its handler (ADR-078)
FAIL  lib/Listener/ExpenseApprovalListener.php (ObjectUpdatedEvent): …
checked 5 post-event registration(s) [diff vs origin/development], 8 out of scope: 2 failure(s)
```

**gate-61 is right, and its finding only became load-bearing because of this PR.** `persist()`
already existed on `development`; `isExpense()` was permanently false, so nothing ever reached it.
Reviving the listener turns two dormant findings into live ones — a synchronous outbound webhook
with retries, plus two object writes, inside another object's write.

Neither remedy the gate offers is available as-is:

- **`ListenerDeferralService` does not exist in pipelinq** — zero hits repo-wide.
- **None of ADR-078's four closed categories** (`realtime`, `sapi-memory`, `cheap-bounded`,
  `correctness`) honestly describes a remote HTTP call with retries on the write path. Writing one
  in anyway would be inventing a fifth category in a docblock, which the checker's own comment
  says needs an ADR amendment.

### 3. Recommendation

**Defer the AP dispatch to a `QueuedJob` before this lands** — the pattern this app already
established in #780 for the walk-in rebalance: `IJobList::add()` + `extends QueuedJob`, which
self-removes after one run and is idempotent for an identical (class, argument) pair. That takes
the webhook and both writes off the request path, satisfies ADR-078 by construction instead of by
annotation, and leaves the in-flight guard in place as defence in depth for the job's own writes.

I have not done that here: it is a separable change with its own tests, and landing this PR as-is
would trade one red gate for another on `development`. **The hold that was placed on this PR is
lifted — the recursion is resolved and proven. It now carries a different, smaller, fully measured
blocker.**

### Also in this push

- Merged `development` in. All fourteen conflicts were the coding-standard reformat from #809
  (same-line braces, tabs, `(string)$x`, `else if` → `elseif`); resolved by taking this branch's
  content and re-running php-cs-fixer, so the result is this branch's semantics in
  `development`'s format. The check that this is what happened: the merged tree differs from
  `development` in exactly the seventeen files this PR touches.
- Fixed a phpmd regression the guard caused (`handle()` at Cyclomatic 10 / NPath 387) by
  **extracting** the five preconditions into `resolveDispatchable()`, not by suppressing the rule.

### Two things worth knowing that came out of verifying this

- **pipelinq is NOT affected by `openregister@c9d96966d`** (which appended `bool $_audit = true` to
  `ObjectService::find()` and broke hermiq's six PHPUnit cells). `tests/bootstrap.php:81`
  pre-declares the local `ObjectService` stub before the Nextcloud bootstrap, so
  `LoyaltyObjectStoreFake extends ObjectService` binds to the stub. Established, not assumed: the
  six PHPUnit cells on the sibling PR started 32 minutes after that commit and all passed, and
  `--filter LoyaltyControllerTest` runs 41 tests / 143 assertions, so it is not a suite that
  silently isn't binding.
- ⚠️ **The same property is a blind spot.** That stub declares
  `find(string $id, string $register = '', string $schema = '')` — it accepts, positionally,
  exactly the call shape that is fatal against the real signature. That is the mechanism behind the
  dead guards in #801/#804 and #805. **The suite is immune to the collaborator changing because it
  never matched the collaborator.** Worth its own change; reddening it will red an unknown number
  of call sites at once.
- **Seven `method_exists($x, 'getUuid')` sites remain outside this PR** (`SlaPolicyController:247`,
  `SlaDeadlineSweepJob:335`, `ActivityTimelineService:497`/`:913`, `SlaEngineService:1435`,
  `EntityActivityService:530`, `InitSlaStatus:271`). Most are probably benign — `getObject()`
  injects `id => uuid` and two of them guard on `$value['id']` first — but they have not been read
  site by site. ⚠️ The `method_exists($x, 'getObject')` sites are a **different** class and must not
  be swept in: `getObject()` is declared concretely, so those probes work.

rubenvdlinde:
### CI has now confirmed the gate-61 cost — the prediction and the run agree

This PR's own gates job:

```
[gate-7]  no-admin-idor: FAIL — 50
[gate-57] orphaned-write-capability: FAIL — 14
[gate-6]  orphan-auth: FAIL — 3
[gate-61] listener-work-placement: FAIL — 2      <- introduced by this PR
[hydra-gates] COVERAGE: 58 of 65 declared gates reported a result (7 not applicable; 58 of 58 applicable gates ran).
```

`gate-61 = 2`, which is what I predicted from the checker before pushing and reproduced through the
full runner locally. Everything else on the PR is green: 34 pass, 4 skipping, and the only red cell
is Hydra Gates.

⚠️ **One thing to read carefully before anyone compares tables.** My local `--full` run of this same
tree reported **six** failing gates; CI reports **four**. Neither is wrong. CI's is a
`pull_request` run and this diff touches no spec file and no page component, so gate-19 (33) and
gate-26 (3) report **NOT APPLICABLE** rather than their real numbers. **This PR's gates cell
therefore looks two gates better than the branch it would land on** — the vacuous-green shape, in a
PR that genuinely changes `lib/`.

So the position is unchanged and now measured twice: **the re-entrancy hold is lifted, and this
should not merge until the AP dispatch is deferred to a `QueuedJob`** (the #780 pattern), because
landing it as-is puts a live outbound webhook with retries on the write path and takes gate-61 red
on `development`.

PR #820, opened 2026-08-13
feat(format): adopt Nextcloud's prettier config, so styles are tabs too

rubenvdlinde:
## The gap this closes

The fleet is tab-indented in **PHP** (php-cs-fixer via `nextcloud/coding-standard`, `setIndent("\t")`) and in **JS/Vue** (`@nextcloud/eslint-config`, `indent: ['error','tab']`).

**CSS and SCSS were enforced by nothing.** `@nextcloud/stylelint-config` carries no indentation rule on the versions the fleet runs — stylelint 15+ dropped stylistic rules. So styles drift away from the `.editorconfig` that says `indent_style = tab`; openregister sits at 391 tab-lines against 511 space-lines, opencatalogi's CSS is 100% spaces.

`prettier` with **`@nextcloud/prettier-config`** (`useTabs: true, tabWidth: 4`) is the tool Nextcloud itself uses for this, and unlike stylelint it covers CSS/SCSS.

| pipelinq styles (`*.css`/`*.scss`, excluding `docs/`) | tab-lines | space-lines |
| --- | --- | --- |
| `development` | 46 | 3 |
| this branch | **49** | **0** |

Follows the pilot, larpingapp#321.

## Why `eslint-config-prettier` is not optional

Measured on this app: with `eslint.config.js` unchanged, prettier's output leaves **625 eslint errors**, every one a formatting rule disagreeing with prettier. Two formatters with overlapping jurisdiction make a repo *unfixable* — the trap this fleet already hit with php-cs-fixer and PHPCS, where `cs:fix` and `phpcs` demanded opposite things and neither could be satisfied.

So `require('eslint-config-prettier')` is spread **last**, after `...conductionVue3Fixes`, where it can only turn rules off. Verified against the resolved config for `src/App.vue`:

- all **21 `vue/no-deprecated-*` rules still present and still ON** — it disables no correctness rule
- `indent` and `quotes` are now severity `0`, owned by prettier's `useTabs: true` instead — the same tab, from the tool that also covers styles

`npm run lint`: **0 errors** (344 pre-existing warnings, unchanged).

## Why the glob is code and styles, not `.`

`nextcloud/forms` runs `prettier --check .`. This fleet cannot. Unrestricted, prettier rewrites markdown and JSON that other tools own:

| path | why that is dangerous |
| --- | --- |
| `openspec/**` | **parsed by the hydra gates** — gate-16 reads `@spec` anchors, gate-19 reads Scenario headings. A formatter and a gate disagreeing about one file is how a gate loses quietly |
| `l10n/**` | written by the translation workflow. This app has `l10n/*.js`, not only `*.json`, so the code glob reaches it — hence `l10n/` in `.prettierignore` |
| `lib/Settings/**.json` | OpenRegister register/schema config, read at install and parsed by the manifest gates |

So the glob is `**/*.{js,ts,vue,css,scss}` — layout-independent, because app styles live under `css/` in some repos and `src/` in others — with build output and the Docusaurus site in `.prettierignore`.

Asserted on this branch: `git diff --name-only | grep -cE '^(openspec|l10n|lib/Settings|docs|\.vscode)/'` is **0**.

## The one non-mechanical change

Three inline **multi-statement** event handlers are extracted to methods (`onCreate`, `onFormCancel` in `PipelineManager.vue`; `onClearSearch` in `ContactImportDialog.vue`). Behaviour is unchanged.

They had to move, and the reason is worth recording. Vue's template compiler only treats a handler as raw *statements* when the expression contains a `;`. Prettier's `semi: false` strips it and reflows the assignments onto separate lines, and the compiler then parses them as one expression and fails:

```
VueCompilerError: Error parsing JavaScript expression: Unexpected token, expected "," (3:6)
```

Measured directly against `@vue/compiler-sfc`:

| handler form | compiles |
| --- | --- |
| `@click="a = true; b = null"` | ok |
| `@click="a = true⏎b = null"` (prettier's output) | **fails** |
| `@click="onCreate"` | ok |

An inline multi-statement handler is an anti-pattern anyway; extracting it is what makes the file stable under the formatter rather than a recurring conflict.

## Verified on this branch

| check | result |
| --- | --- |
| `npm run format` | clean |
| `npm run lint` | **exit 0**, 0 errors |
| `npm run stylelint` | exit 0 |
| `npm run test:unit` | **45/45** |
| `npm run build` | webpack compiles |
| non-code paths touched | **0** |

The resolved config was asserted *before* formatting, because a prettier run that finds no config silently falls back to defaults and writes **double quotes**, which `@nextcloud/prettier-config` forbids — and it looks like a successful run. It reads `useTabs=true tabWidth=4 singleQuote=true semi=false printWidth=85`.


rubenvdlinde:
## CI: what is mine and what is not

Every job below was compared against the **same job on `development`** — the newest `push` run of `Code Quality`, not a status flag.

### Hydra Gates

| gate | on `development` | on this PR |
| --- | --- | --- |
| gate-57 orphaned-write-capability | FAIL 14 | FAIL 14 |
| gate-6 orphan-auth | FAIL 3 | FAIL 3 |
| gate-7 no-admin-idor | FAIL 50 | FAIL 50 |
| gate-16 spec-coverage | **PASS** | **FAIL — 468** |

**gate-16 is the only new failure, and it is a false positive.** The gate inspects frontend methods as well as PHP ones. Its recent normalisation fix (`.github#399`) deliberately does not normalise **JS** re-wraps, because in JavaScript a re-wrap can interact with automatic semicolon insertion and array elision. Prettier re-wraps `.vue`/`.js` heavily, so every touched frontend method reads as "changed" and the gate asks for an `@spec` tag on it.

Nothing here has been annotated to satisfy it and the glob has not been narrowed to dodge it — adding `@spec` tags to methods nobody changed would bake the defect in. The gate fix lands first; this PR waits for it.

Every other gate carries an **identical count on both sides**.

PR #827, opened 2026-08-14
fix(security): gate the CRM endpoints — gate-7 from 50 findings to 1

rubenvdlinde:
**gate-7 `no-admin-idor`: 50 → 1**, measured with the real gate after each batch. The one remaining is a false positive, explained below.

Every one of these was an `#[NoAdminRequired]` endpoint reachable by **any authenticated account on the instance** — no per-object guard, no group check.

### The one worth reading first

`BrpController::validate` validates a **BSN** — the Dutch citizen service number, special-category personal data under the AVG. It was open to any logged-in user. That is the most sensitive endpoint in the set and it had nothing in front of it.

### Approach, and why it isn't per-object ownership

23 of 27 schemas have **no ownership field**, so an ownership guard was literally unwritable for most of these. The answer is group-based: the existing `ObjectOwnerAccessPolicy` gained a configurable `crm_group` alongside its existing privileged groups.

I first built a new `CrmAccessPolicy` and deleted it — it duplicated a class that already existed. Consolidating onto `ObjectOwnerAccessPolicy` is commit `f2e2e94`.

**Fail-closed by design:** until an admin sets `crm_group` to a group with members, only Nextcloud admins reach the CRM. That was explicitly accepted.

### Two places the guard is deliberately *not* what you'd expect

- **`Customer360Controller`** — the guard is added **in front of** `canReadClient()`, which is left untouched. That helper calls `ObjectService::find()` positionally and catches `Throwable` to return false, and behind it sits `return $object !== null` — an *existence* test. Repairing it would turn a dead endpoint into a live IDOR over every client record.
- **Blast/Segment** — the guard started in a `requireUser()` helper. gate-7 doesn't accept that (a helper returning null isn't a guard), and worse, keeping both would have made the body guard **unreachable**. Moved into the bodies.

### The remaining gate-7 finding is a false positive

`PortalController::availability`, rule `publicpage-unscoped-object-lookup` — a genuinely public booking endpoint, so an auth guard would be wrong. And the lookup is scoped: it delegates to `BookingService::getAvailableSlots`, which calls `find(id:, register:, schema:)` with named arguments against this app's own register, failing closed when unconfigured. gate-7 doesn't follow the controller→service delegation. Contorting the app to satisfy it would be making a cell green over a defect that isn't there.

### State

Merged `development` in — **0 behind, zero conflicts**, and gate-7 still reports 1 and gate-16 now PASSes on the merged result. 23 files, 728 insertions.

Pre-existing and untouched: gate-6 (3 orphan-auth) and gate-57 (14 orphaned write-capability).


rubenvdlinde:
**Ready for review — the suite is green: 2210 tests, 0 failures** (development's baseline is 2200; this branch adds 10).

### Correction to my earlier notes on this PR

I twice mis-diagnosed the residual failures. First as a **semantic merge conflict** — that the english-vocabulary migration had renamed `$gekoppeldRequest` and git had merged it cleanly into a broken state. Then as a knock-on from a missed constructor. **Both wrong.**

The truth is the opposite: **this branch is *ahead* on the migration.** Its `BsnAuditService` already uses `$action` / `$gekoppeldRequest` / `$actorRole`; `development` still has `$actie` / `$gekoppeldVerzoek` / `$actorRol`. Nothing was broken by a merge.

The error came from **my own verification harness**. I test by copying files into the shared dev checkout, and I was copying only `lib/Controller` and `tests/Unit/Controller` — so the container ran *this branch's* controllers against *development's* services and tests. Two codebases in one tree, with nothing announcing it. Copying `lib/` and `tests/` whole took 21 failures to zero **without one line of production code changing**.

### One real bug fixed along the way

`ActivityTimelineController::objectAccessible()` called `jsonSerialize()` unconditionally on whatever the container returned. That seam is untyped — resolved *by string* — so on an array it fataled into the catch and turned a **type error into a silent "deny"**: a 404 that read as an authorization decision. Now handles both shapes.

### One thing worth keeping in review

`SemanticHandoffControllerTest`'s `isInGroup => false` default is **load-bearing in both directions**, and I verified that rather than assuming: flip it to true and the ownership tests sail past the guard into service calls they assert are never made; leave it false and seven business-logic tests get a 403 before the behaviour they're about. The seven now request privilege individually via `withPrivilegedCaller()`; the default stays.

PR #833, opened 2026-08-14
refactor(deps): inject OpenRegister instead of looking it up (ADR-083)

rubenvdlinde:
65 file(s) reached OpenRegister through `$this->container->get(...)` on an **unconditional** path — no availability check, no degrading catch. The dependency was announced nowhere: not in the constructor, not in the `use` block, not in any type. It appeared mid-method, as a **string**.

Now constructor-injected and typed. Behaviour is unchanged — same object, same container, resolved at construction instead of at first use. `ContainerInterface` is dropped only where nothing else used it.

**Deliberately not converted**, because they are correct as written (ADR-083 rule 1's exception): lookups behind `isInstalled()`/`getInstalledApps()`, and lookups whose `catch` degrades rather than rethrows. Converting those would make the service unconstructable without OpenRegister and turn a clean message into a 500.

Verified per file: `php -l` clean, and gate-66's lookup check reports **zero** remaining findings for each file changed.

gate-66 for this app: **78 → 8**.

### DIFF
--- a/lib/BackgroundJob/BrpMonitorJob.php
+++ b/lib/BackgroundJob/BrpMonitorJob.php
@@ -34,9 +34,9 @@ use OCP\BackgroundJob\TimedJob;
 use OCP\IAppConfig;
 use OCP\IGroupManager;
 use OCP\Notification\IManager as INotificationManager;
-use Psr\Container\ContainerInterface;
 use Psr\Log\LoggerInterface;
 use Throwable;
+use OCA\OpenRegister\Service\ObjectService;
 
 /**
  * BRP availability + SLA monitor.
@@ -67,10 +67,10 @@ class BrpMonitorJob extends TimedJob {
 	public function __construct(
 		ITimeFactory $time,
 		private IAppConfig $appConfig,
-		private ContainerInterface $container,
 		private IGroupManager $groupManager,
 		private INotificationManager $notificationManager,
 		private LoggerInterface $logger,
+		private readonly ObjectService $objectService,
 	) {
 		parent::__construct(time: $time);
 		$this->setInterval(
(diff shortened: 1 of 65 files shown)


rubenvdlinde:
Marked draft — **not mergeable as it stands.**

Removing `ContainerInterface` from these constructors shifts every *positional* argument after it, so unit tests that construct these services pass the wrong mock into the next slot:

```
TypeError: __construct(): Argument #2 ($appManager) must be of type
IAppManager, MockObject_ContainerInterface given
```

**75 test file(s) in this repo construct the classes this PR changes.** They all need updating to the new signature before this can go in — that is the real cost of the conversion, and it is not optional: a double pinned to a signature that moved reports a defect that does not exist.

The `lib/` side is complete and verified — `php -l` clean per file, and gate-66's lookup check reports zero remaining findings for every file changed. What is missing is the test half.

Kept open rather than closed so the `lib/` work is not lost. docudesk ([#461](https://github.com/ConductionNL/docudesk/pull/461)) shows the shape of the fix on a 2-file example.

rubenvdlinde:
## Blocked on a missing prerequisite, not on this diff

Piloting the test fixes surfaced why every rollout PR goes red the same way, and it is not the conversion:

**OpenRegister is not autoloadable in unit tests.** This repo's `tests/bootstrap.php` declares stubs for OpenRegister *events and interfaces* precisely because the real classes are absent in the standalone composer environment. Tests that need `ObjectService` therefore either wire it through a `ContainerInterface` mock (untyped, so anything passes) or declare a **local stub interface** in the test file.

Constructor-injecting the concrete `OCA\OpenRegister\Service\ObjectService` type removes that escape hatch: PHPUnit cannot mock a class it cannot load.

**Only `openconnector` ships `tests/stubs/OCA/OpenRegister/Service/ObjectService.php`.** The other seven apps do not, which is exactly the set of red PRs.

### So the recipe has a step nobody had written down

1. add an `ObjectService` test stub + autoload it from `tests/bootstrap.php`
2. transform `lib/` ← **this PR, complete and verified**
3. update the test constructions

### And step 1 is a decision, not a chore

openconnector's stub declares **10 of the real class's 59 methods**. Vendoring seven more copies means seven doubles drifting away from a class none of them own — and a stale double on this fleet has already reported a security hole that did not exist.

The alternative is OpenRegister publishing a contract (an interface, or a `tests/stubs` package consumers require) so there is one definition rather than eight.

**Left as a draft deliberately.** `lib/` is complete: `php -l` clean per file, gate-66 lookup findings zero for every file changed. It should land once step 1 is settled — not before.

PR #842, opened 2026-08-16
fix(build): compare the sibling nc-vue to the declared range, not its peer vue

rubenvdlinde:
The guard here was opt-in and it did check something — but it asked the wrong
question, so it answered "safe" about an unsafe sibling.

It read the sibling's `peerDependencies.vue` and refused anything that was not
`3.x`. The sibling declares `vue: ^3.5.0`. It IS a Vue 3 library — and it is
2.0.5, against a declared ^2.3.0. Being on the Vue 3 line and being the
version this app asked for are different properties, and only the second one
makes the sibling safe to alias in. The check passed and the wrong library
went into the build.

Now the sibling must satisfy the declared RANGE, checked with semver, failing
CLOSED if the check itself cannot run.

One behaviour change worth calling out: a rejected sibling now warns and falls
back to the pinned npm package instead of throwing. Refusing the sibling
yields a complete, correct build, so a throw turned a safe fallback into a
broken build for no gain.

Measured: USE_LOCAL_LIB unset builds clean with the sibling unused;
USE_LOCAL_LIB=true builds clean AND prints
`IGNORING sibling @conduction/nextcloud-vue@2.0.5` — the guard shown refusing,
which the old check never did against this sibling.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -98,21 +98,47 @@ webpackConfig.entry = {
 // Vue 3 app. Opt IN explicitly (USE_LOCAL_LIB=true) and hard-fail if the local
 // tree is not on the Vue 3 major.
 const localLib = path.resolve(__dirname, '../nextcloud-vue/src')
-const useLocalLib = process.env.USE_LOCAL_LIB === 'true' && fs.existsSync(localLib)
+let useLocalLib = process.env.USE_LOCAL_LIB === 'true' && fs.existsSync(localLib)
 if (useLocalLib) {
-	const localPkg = JSON.parse(
-		fs.readFileSync(
-			path.resolve(__dirname, '../nextcloud-vue/package.json'),
-			'utf8',
-		),
-	)
-	const localVue =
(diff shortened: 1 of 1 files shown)


rubenvdlinde:
Merging with admin. The `quality / Hydra Gates` failure is not introduced by this PR.

This branch changes **`webpack.config.js` only**. The gates that fail are `gate-7` (controller methods with `NoAdminRequired` and no per-object guard) and `gate-19` (scenarios missing `@e2e`) — both scanned full-repo, neither reachable from a webpack build config.

The reason it *looked* new is worth recording, because it is the second time tonight the same shape has appeared: on the base commit, `quality / Hydra Gates` is `failure`. A delta rule that asks "is this check failing on head and not failing on base?" reads a base with no verdict exactly like a base that passed. It is neither. Gates also run full-repo rather than diff-scoped, so any finding on `development` reddens every PR opened against it.

The underlying findings are real debt for `development` — they are simply not this branch's doing, and this branch cannot fix or worsen them.

